PR #47, opened 2020-04-03
Sync will invite, uninvited, remove and ignore users

huwr:
Previously, we were only looking at the users list, and inviting people who were not on it. This causes trouble when run twice. If we had invited someone, but they have not yet accepted their invitation, the code would attempt to invite them again. That is not allowed by the API. It would return an error and the program would end.

The reason the code would attempt to invite them twice is because it only checked the Users API endpoint. It also needed to check the Pending Invitations to see who has been invited.

Now, we look at not only who is in the users list, but also who has been invited. This means you can run the program many times with the same list, and it will have the [same effect](https://en.wikipedia.org/wiki/Idempotence). A user who has already invited won't be invited a second time.

For everyone in the input list, we check if they are in either the existing users or the invited users lists. If they are, we can ignore that entry. If they are not, we invite them.

For everyone in the existing users list that we want to remove, there are two things we might have to do: remove them from the users list, or remove them from the invited users list. Since the users and invited users are separate API calls there are two separate operations for deleting users or uninviting them. 

This does everything in https://github.com/ittybittyapps/appstoreconnect-cli/issues/5 except for _modification_. If a user in the input file has a role that they do not have in the API, we should be adding that role. 

## 📝 Summary of Changes

- Make sync remove and ignore invitations

## ⚠️ Items of Note

Please don't do this on the IBA users list. :+1:

### 💁 Example

Say this is our current user list:

| User | email | Accepted Invation? |
|------| ------| ------|
| Carol | carol@example.com | true |  
| Edith | edith@example.com | false |  
| Fran | fran@example.com | true |  

We wish to remove `Edith`, who never accepted the invitation, and `Fran`, who no longer works here.

Start by grabbing all my existing users. We'll use this as our input file: 

```
./appstoreconnect-cli users list --output-format json > users.json
```

I'll add `Alice`, `Bob` and `Dave` to `users.json`, and remove `Edith` and `Fran`.

```
./appstoreconnect-cli users sync --input-format json users.json
invited alice@example.com
invited bob@example.com
ignored carol@example.com
invited dave@example.com
uninvited edith@example.com
removed fran@example.com
```

If I run that a second, third, or fourth time it becomes:

```
./appstoreconnect-cli users sync --input-format json users.json
ignored alice@example.com
ignored bob@example.com
ignored carol@example.com
ignored dave@example.com
```

All the entries are now `ignored`. They already exist as users or have pending invitations.

### 🔨 How To Test

Please do not do this on IBA's users list. I recommend doing it on your own App Store Connect account. Remember _uninviting_ and _removing_ are separate operations.

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/Users/GetUserInfoCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Users/GetUserInfoCommand.swift
@@ -29,7 +29,7 @@ struct GetUserInfoCommand: ParsableCommand {
         let request = APIEndpoint.users(filter: filters)
 
         _ = api.request(request)
-            .map(User.fromAPIResponse)
+            .map(Array<User>.init)
             .sink(
                 receiveCompletion: Renderers.CompletionRenderer().render,
                 receiveValue: Renderers.ResultRenderer(format: outputFormat).render
--- a/Sources/AppStoreConnectCLI/Commands/Users/Invitations/CancelUserInvitationsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Users/Invitations/CancelUserInvitationsCommand.swift
@@ -21,7 +21,7 @@ struct CancelUserInvitationsCommand: ParsableCommand {
 
         let cancelInvitation = { api.request(APIEndpoint.cancel(userInvitationWithId: $0)) }
 
-        _ = try api
+        _ = api
             .invitationIdentifier(matching: email)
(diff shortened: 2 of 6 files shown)


orj:
Sorry @huwr, gonna close this and look at it again in future.

huwr:
Haha that's okay! A stub is implemented, but has a few problems. I think there would be a different way to implement it now, too, with all the changes made to the codebase since this PR was opened.

Issue #45, opened 2020-04-02
Commands to list, modify and delete TestFlight Beta Testers


DechengMa:
# 🗣 Context

We have issue #12, and creating beta testers is one of the main features of AppStoreConnect API.
But it can't cover all the use cases, and we want to do more about TestFlight testers
# 💬 Narrative

As a user
I want to be able to
- list all the beta testers for an app,
- read one or more beta tester's info
- removing one or more beta tester

So that I can have better management of all the beta testers   
Covering APIs
1. [list](https://developer.apple.com/documentation/appstoreconnectapi/list_beta_testers)
2. [info](https://developer.apple.com/documentation/appstoreconnectapi/read_beta_tester_information)
3. [delete](https://developer.apple.com/documentation/appstoreconnectapi/delete_a_beta_tester)

# ✅ Acceptance Criteria

- I can use the command `list` to see a list of all the beta testers of an app, by inputting app's `bundleId`
- I can input beta testers' emails, to get specific beta testers info
- I can input beta testers' emails to delete one or more beta testers 

# 🚫 Out of Scope

This Issues not relevant to Beta Groups, we will have another issue for Beta Groups

Issue #41, opened 2020-04-02
Task: Create a Domain layer


adamjcampbell:
# ❕ Problem Statement

We currently call directly into `AppStoreConnect_Swift_SDK` and couple ourselves to it by directly making use of it from our commands

We mix and match whether we use `AppStoreConnect_Swift_SDK` models (with extensions) or if we translate into our own Models

It's reaching the point where we want our own consistent Model layer and to abstract the knowledge of `AppStoreConnect_Swift_SDK` away from our commands

# 💬 Task Description

- Create Models for use in all our commands
- Create an API Client that calls through to `AppStoreConnect_Swift_SDK` and translates the library model types to our model types

# 👩‍🔧 Technical Design Notes

We should structure our models with consideration for table formats, some models we use are quite nested
If we are to support pretty output tables and `csv` we need a flat structure that can print models row by row

# 🤝 Relationships

#24 
This work is needed to output CSV correctly as CSVDecoder does not support types that are nested more than two levels.

adamjcampbell:
@orj shall we close this now that we have an established pattern in #80 ?

adamjcampbell:
Superseded by #177

PR #58, opened 2020-04-07
Implement ListBetaTestersCommand

DechengMa:
Implement List Beta Tester Command, four ways of listing are supported, this PR is related to #45 

# 📝 Summary of Changes

- Update `appstoreconnect-swift-sdk` to 1.0.3
- Create `ListBetaTestersCommand`, user can list beta testers who are related to 
1. an app,
3. a beta group
4. all, if there is no input
- Create `ListBetaTestersByBuilds`, user can list beta testers who were assigned to one or more specific builds

# Notes
1. Can't input both group name and bundle, API will return `PARAMETER_ERROR.INVALID` and `Only one relationship filter can be applied`, and an error will be thrown
2. For listing the by beta group name, there will be no `include` with `relationship` data in response. I already filed feedback to Apple, haven't got something back now.

# 🧐🗒 Reviewer Notes

## 💁 Example

`testflight betatesters list [--bundle-id <bundle-id>] [--beta-group-name <beta-group-name>] [--build-id <build-id>] [--related-resources-limit <related-resources-limit>]`

`testflight betatesters listByBuilds <bundle-id> [--pre-release-versions <pre-release-versions> ...] [--versions <versions> ...]`

## 🔨 How To Test

`swift run asc testflight betatesters list --group-name yourGroupName`

`swift run asc testflight betatesters listByBuilds com.example.roadrunner --pre-release-versions 1.0.0 --versions 1 `

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
@@ -81,7 +81,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
         }
 
         _ = request
-            .map { BetaTester.init($0.data, $0.included) }
+            .map { BetaTester($0.data, $0.included) }
             .renderResult(format: common.outputFormat)
     }
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
@@ -81,7 +81,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
         }
 
         _ = request
-            .map { BetaTester.init($0.data, $0.include) }
+            .map { BetaTester.init($0.data, $0.included) }
             .renderResult(format: common.outputFormat)
     }
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
@@ -84,7 +84,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
             .map(\.data)
             .sink(
                 receiveCompletion: Renderers.CompletionRenderer().render,
-                receiveValue: Renderers.ResultRenderer(format: common.outputFormat).render
+                receiveValue: { _ in }
             )
     }
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
@@ -81,10 +81,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
         }
 
         _ = request
-            .map(\.data)
-            .sink(
-                receiveCompletion: Renderers.CompletionRenderer().render,
-                receiveValue: { _ in }
-            )
+            .map { BetaTester.init($0.data, $0.include) }
+            .renderResult(format: common.outputFormat)
     }
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
@@ -29,14 +29,11 @@ struct ListBetaTesterByBuildsCommand: CommonParsableCommand {
     )
     var versions: [String]
 
-    private enum CommandError: Error, LocalizedError {
-        case noAppsFound(bundleId: String)
+    private enum CommandError: LocalizedError {
         case noBuildsFound(preReleaseVersions: [String], versions: [String])
 
-        var errorDescription: String? {
+        var failureReason: String? {
             switch self {
-            case .noAppsFound(let bundleId):
-                return "No apps were found matching \(bundleId)"
             case .noBuildsFound(let preReleaseVersions, let versions):
                 return "No builds were found matching preReleaseVersions \(preReleaseVersions) and versions \(versions)"
             }
@@ -48,11 +45,7 @@ struct ListBetaTesterByBuildsCommand: CommonParsableCommand {
 
         _  = api
             .getAppResourceIdsFrom(bundleIds: [bundleId])
-            .flatMap { [versions, preReleaseVersions, bundleId] appIds -> AnyPublisher<BuildsResponse, Error> in
-                guard !appIds.isEmpty else {
-                    let error = CommandError.noAppsFound(bundleId: bundleId)
-                    return Fail(error: error as Error).eraseToAnyPublisher()
-                }
+            .flatMap { [versions, preReleaseVersions] appIds -> AnyPublisher<BuildsResponse, Error> in
 
                 var filters: [ListBuilds.Filter] = [.app(appIds)]
 
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
@@ -42,7 +42,7 @@ struct ListBetaTesterByBuildsCommand: CommonParsableCommand {
                 }
 
                 if !preReleaseVersions.isEmpty {
-                    filters.append(.version(preReleaseVersions))
+                    filters.append(.preReleaseVersionVersion(preReleaseVersions))
                 }
 
                 return api.request(APIEndpoint.builds(filter: filters))
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
@@ -72,7 +72,7 @@ struct ListBetaTesterByBuildsCommand: CommonParsableCommand {
                 return api.request(endpoint).eraseToAnyPublisher()
             }
             .map { response in
-                response.data.map { BetaTester.init($0, response.included) }
+                response.data.map { BetaTester($0, response.included) }
             }
             .renderResult(format: common.outputFormat)
     }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
@@ -1,0 +1,64 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct ListBetaTesterByBuildsCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "listByBuilds",
+        abstract: "List beta testers who were specifically assigned to one or more builds"
+    )
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Argument(help: "The bundle ID of an application. (eg. com.example.app)")
+    var bundleId: String
+
+    @Option(
+        parsing: .upToNextOption,
+        help: "The pre-release version number of this build. (eg. 1.0.0)"
+    )
+    var preReleaseVersions: [String]
+
+    @Option(
+        parsing: .upToNextOption,
+        help: "The version number of this build. (eg. 1)"
+    )
+    var versions: [String]
+
+    func run() throws {
+        let api = try makeClient()
+
+        _  = api
+            .getAppResourceIdsFrom(bundleIds: [bundleId])
+            .flatMap { [versions, preReleaseVersions] appIds -> AnyPublisher<BuildsResponse, Error> in
+                var filters: [ListBuilds.Filter] = [.app(appIds)]
+
+                if !versions.isEmpty {
+                    filters.append(.version(versions))
+                }
+
+                if !preReleaseVersions.isEmpty {
+                    filters.append(.version(preReleaseVersions))
+                }
+
+                return api.request(APIEndpoint.builds(filter: filters))
+                    .eraseToAnyPublisher()
+            }
+            .flatMap {
+                api.request(APIEndpoint.betaTesters(
+                        filter: [.builds($0.data.map(\.id))],
+                        include: [.apps, .betaGroups]
+                    )
+                )
+            }
+            .map { response in
+                response.data.map { BetaTester.init($0, response.included) }
+            }
+            .renderResult(format: common.outputFormat)
+    }
+
+}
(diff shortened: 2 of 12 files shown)


Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,91 @@
...
+
+struct ListBetaTestersCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "list",
+        abstract: "List beta testers")
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Option(default: "",
```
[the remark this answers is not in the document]
DechengMa: if it's optional, for `case let(bundleId, _, _) where !bundleId.isEmpty:`  I have to change to 
```
case let (bundleId, _, _) where bundleId != nil:
    self = .listByApp(bundleId: bundleId!)
```
because I don't know how to use `if let` or `guard` to removing the force unwrapping`!`?

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,91 @@
...
+        case all
+        case listByApp(bundleId: String)
+        case listByGroup(betaGroupName: String)
+        case listByBuild(buildId: String)
+
+        typealias ListOptions = (bundleId: String, betaGroupName: String, buildId: String)
+
+        init(options: ListOptions) {
+            switch (options.bundleId, options.betaGroupName, options.buildId) {
+                case (let bundleId, _, _) where !bundleId.isEmpty:
```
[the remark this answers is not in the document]
adamjcampbell: Maybe I'm the odd one out but I have always preferred putting the variable binding inside the match expression

- All your matches line up
```swift
case (let string):
  break
case ("specificString"):
  break
```
- You can choose the mutability per binding
```swift
case (let foo, var bar):
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,91 @@
...
+            case .listByBuild(let buildId):
+                request = api.request(APIEndpoint.betaTesters(
+                        filter: [ListBetaTesters.Filter.builds([buildId])]
+                    ))
+                    .eraseToAnyPublisher()
+
+        }
+
+        _ = request
+            .map(\.data)
```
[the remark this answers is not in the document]
adamjcampbell: I believe we're allowing output as:
- pretty table
- json
- yaml

With the idea being that outputted json / yaml could be consumed by or piped into some other tool

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,91 @@
...
+                request = try api.betaGroupIdentifier(matching: betaGroupName)
+                    .flatMap {
+                        api.request(APIEndpoint.betaTesters(
+                            filter: [ListBetaTesters.Filter.betaGroups([$0])]
+                        ))
+                    }
+                    .eraseToAnyPublisher()
+            case .listByBuild(let buildId):
+                request = api.request(APIEndpoint.betaTesters(
+                        filter: [ListBetaTesters.Filter.builds([buildId])]
```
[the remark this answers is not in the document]
adamjcampbell: We should be able to use a combination of the bundle id and the build number I think

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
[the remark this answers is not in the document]
adamjcampbell: I refer to googles style guide for swift a lot as it's very explicit
Not sure what we want to use going forward for this project
They have a section on pattern matching: https://google.github.io/swift/#pattern-matching

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
```diff
@@ -84,7 +84,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
             .map(\.data)
             .sink(
                 receiveCompletion: Renderers.CompletionRenderer().render,
-                receiveValue: Renderers.ResultRenderer(format: common.outputFormat).render
+                receiveValue: { _ in }
```
DechengMa, line 87: Following what we have for other Commands (`CancelUserInvitationsCommand` `DeleteBetaTesterCommand` etc), no error or no result means success.

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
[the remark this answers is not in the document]
DechengMa: > We should be able to use a combination of the bundle id and the build number I think

Just found we can input the `bundled`, `prerelaseVerison`  and `version` to achieve that. But the `AppStoreConnect_Swift_SDK` has a naming Error according to the fetch Build feature which needs to be solved first

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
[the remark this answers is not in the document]
orj: We'll I think now would be a great time to define a style guide for the project. IBA has one, https://github.com/ittybittyapps/engineering/blob/master/Standards/Swift-Style-Guide/Swift-Style-Guide.md.  If you think that could be improved based on other sources of guidance then I'm all 👂👂

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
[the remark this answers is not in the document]
orj: Yep.  But the data should be the same either way.  Ie, what gets output as YAML should be the same as the JSON and the table. In this case the JSON was just the raw response data.  We should be mapping that to some sort of model of our own that represents the results of our API queries.  As the result might actually be dependant on multiple queries against the underlying API.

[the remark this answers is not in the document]
adamjcampbell: Hmm, maybe this ties into #41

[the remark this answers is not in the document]
DechengMa: Yeh, it's better to do the domain layer things earlier. Is there any great template or example for creating a domain layer that I can follow?

orj:
I notice now that this command is very slow.

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,97 @@
...
+                request = api.request(APIEndpoint.betaTesters(
+                        filter: [ListBetaTesters.Filter.builds([buildId])],
+                        include: includes
+                    ))
+                    .eraseToAnyPublisher()
+        }
+
+        _ = request
+            .map(\.data)
+            .flatMap { api.fromAPIBetaTesters(betaTesters: $0) }
```
orj, line 91: This flatMap is going to create a whole load of redundant requests.  Which is probably why this command can take so long now if you have a bunch of beta testers. With an account with 50 beta testers it takes nearly 30seconds to run.

DechengMa: Yes, because the response of getting beta tester will only return Ids of Apps and Ids of Groups, in order to get the name and bundle Id of the group and App, I use the Id to fetch Groups/Apps respectively.

orj:
@DechengMa My recommendation on attacking this task would be:

First fetch all the beta testers (all pages). In my initial testing one page of 50 users takes <2 seconds to fetch.

If you `include` `apps`, `builds` and `betaGroups` that adds another second to the request.
However the default limit on included resources is about 10 results per beta tester.  For large orgs this might be a problem.  You can increase these limits to 50 through query options though. So you could give that power to the user.

If you include relationship data in the request via `include` you can work out the app names and beta group names easily from the included resources.  You don't need to do further queries against the API.

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,97 @@
...
+struct ListBetaTestersCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "list",
+        abstract: "List beta testers")
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Option(help: "The bundle ID of an application. (eg. com.example.app)")
+    var bundleId: String?
```
orj, line 17:
```suggestion
    var filterBundleId: String?
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    @Option(help: "The name of the beta group")
+    var betaGroupName: String?
```
orj, line 20:
```suggestion
    var filterBetaGroupName: String?
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    @Option(help: "The ID of one build of an application")
+    var buildId: String?
```
orj, line 23:
```suggestion
    var filterBuildId: String?
```

orj: Also, you should be able to specify these filter options in combinations.  The code below seems to assume you only use one at a time.

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/CreateBetaTesterCommand.swift
```diff
@@ -81,10 +81,7 @@ struct CreateBetaTesterCommand: CommonParsableCommand {
         }
 
         _ = request
-            .map(\.data)
-            .sink(
-                receiveCompletion: Renderers.CompletionRenderer().render,
-                receiveValue: Renderers.ResultRenderer(format: common.outputFormat).render
-            )
+            .map { BetaTester.init($0.data, $0.included) }
```
adamjcampbell, line 84:
```suggestion
            .map { BetaTester($0.data, $0.included) }
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTesterByBuildsCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+                    return Fail(error: error as Error).eraseToAnyPublisher()
+                }
+
+                let endpoint = APIEndpoint.betaTesters(filter: [.builds(buildResponse.data.map(\.id))],
+                                                       include: [.apps, .betaGroups])
+
+                return api.request(endpoint).eraseToAnyPublisher()
+            }
+            .map { response in
+                response.data.map { BetaTester.init($0, response.included) }
```
adamjcampbell, line 75:
```suggestion
                response.data.map { BetaTester($0, response.included) }
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,106 @@
...
+                        ))
+                    }
+                    .eraseToAnyPublisher()
+            case .listByAppAndGroup:
+                throw ListBetaTesterError.multipleFilters
+        }
+
+        _ = request
+            .map { response in
+                response.data.map { BetaTester.init($0, response.included) }
```
adamjcampbell, line 102:
```suggestion
                response.data.map { BetaTester($0, response.included) }
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -0,0 +1,106 @@
...
+                    .flatMap {
+                        api.request(APIEndpoint.betaTesters(
+                            filter: [.betaGroups([$0])],
+                            include: includes,
+                            limit: limits
+                        ))
+                    }
+                    .eraseToAnyPublisher()
+            case .listByAppAndGroup:
+                throw ListBetaTesterError.multipleFilters
```
adamjcampbell, line 97: Let's make every all paths go through the completion handler
Instead of throwing from here we can set the request to be a `Fail` publisher rather than exiting early

Sources/AppStoreConnectCLI/Model/BetaTester.swift
```diff
@@ -5,74 +5,91 @@ import Combine
...
-            TextTableColumn(header: "First Name"),
-            TextTableColumn(header: "Last Name"),
-            TextTableColumn(header: "Invite Type")
-        ]
+    init(email: String?,
+         firstName: String?,
+         lastName: String?,
+         inviteType: BetaInviteType?,
+         betaGroups: [BetaGroup]?,
+         apps: [App]?) {
```
adamjcampbell, line 21:
```suggestion
    init(
        email: String?,
        firstName: String?,
        lastName: String?,
        inviteType: BetaInviteType?,
        betaGroups: [BetaGroup]?,
        apps: [App]?
    ) {
```

Sources/AppStoreConnectCLI/Model/BetaTester.swift
```diff
@@ -5,74 +5,91 @@ import Combine
...
+            return nil
+        }
 
-    var betaGroupsIds: [String] {
-        return relationships?.betaGroups?.data?.compactMap { $0.id } ?? []
-    }
+        let betaGroupsInTester = relationships?.betaGroups?.data?.compactMap { group -> [BetaGroup]? in
+                includedBetaGroups?.filter { $0.id == group.id }
+            }
+            .flatMap { $0 }
```
adamjcampbell, line 51:
```suggestion
        let betaGroupsInTester = relationships?.betaGroups?.data?
            .compactMap { group -> [BetaGroup]? in
                includedBetaGroups?.filter { $0.id == group.id }
            }
            .flatMap { $0 }
```

Sources/AppStoreConnectCLI/Model/BetaTester.swift
```diff
@@ -5,74 +5,91 @@ import Combine
...
 
-    var buildsIds: [String] {
-        return relationships?.builds?.data?.compactMap { $0.id } ?? []
+        let appsInTester = relationships?.apps?.data?.compactMap { app -> [AppStoreConnect_Swift_SDK.App]? in
+                includedApps?.filter { app.id == $0.id }
+            }
```
adamjcampbell, line 55:
```suggestion
        let appsInTester = relationships?.apps?.data?
            .compactMap { app -> [AppStoreConnect_Swift_SDK.App]? in
                includedApps?.filter { app.id == $0.id }
            }
```

Sources/AppStoreConnectCLI/Model/BetaTester.swift
```diff
@@ -5,74 +5,90 @@ import Combine
 import Foundation
 import SwiftyTextTable
 
-extension BetaTester: ResultRenderable { }
+struct BetaTester: ResultRenderable {
+    let email: String?
```
adamjcampbell, line 9: It would be nice if we could define what was optional and what was non optional but we can do a pass over that in a different PR

adamjcampbell:
@orj I dismissed your review so we could get this in, it'd be nice to get it retested with your dataset and raise an issue if it's still not as performant as you expect.

Issue #69, opened 2020-04-09
Bug: Users list --limit option doesn't work


Nafisa2020:
# 🌎 Environment

 - OS: macOS 
 - Version/Release: 10.15.2

# 💬 Description
The limit option for users list doesn't work. It returns the entire list.
 

# 🦶 Reproduction Steps

Steps to reproduce the behavior, provide an example of the command line issued:
1. Run this `swift run asc users list --limit 5`
 
# 🤔 Expected Results
Should return only 5 users if the limit is set as 5

# 😲 Actual Results

It returns the entire list

PR #70, opened 2020-04-09
Correctly implement limit visible apps for list user commands

Nafisa2020:
# 📝 Summary of Changes

Right now the sdk is only supporting limit for visible apps. So for the time being 
- Renamed limit optional argument to limitVisibleApps
- Passed the value of the limitVisibleApps  to the limit parameter of the request for ListUsersCommand
- Changed the wordings of the option help text for ListUsersCommand and ListUserInvitationsCommand

More changes can be added when the sdk supports limit for users

# 🔨 How to test
swift run asc users list --limit-visible-apps 2

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
@@ -13,7 +13,7 @@ struct ListUserInvitationsCommand: CommonParsableCommand {
     @OptionGroup()
     var common: CommonOptions
 
-    @Option(help: "Limit the number visible apps to return (maximum 200).")
+    @Option(help: "Limit the number visible apps to return (maximum 50).")
     var limit: Int?
 
     @Option(parsing: .upToNextOption, help: "Filter the results by the specified username")
--- a/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
@@ -43,7 +43,7 @@ struct ListUserInvitationsCommand: CommonParsableCommand {
         let api = try makeClient()
 
         let endpoint = APIEndpoint.invitedUsers(
-            limit: limitVisibleApps.map { [ListInvitedUsers.Limit.visibleApps($0)] } ?? [],
+            limit: limitVisibleApps.map { [ListInvitedUsers.Limit.visibleApps($0)] } ,
             filter: filters
         )
 
--- a/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
@@ -14,7 +14,7 @@ struct ListUserInvitationsCommand: CommonParsableCommand {
     var common: CommonOptions
 
     @Option(help: "Limit the number visible apps to return (maximum 50).")
-    var limit: Int?
+    var limitVisibleApps: Int?
 
     @Option(parsing: .upToNextOption, help: "Filter the results by the specified username")
     var filterEmail: [String]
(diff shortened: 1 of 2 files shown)


adamjcampbell:
@Nafisa2020 I believe that there's actually a naming error in the SDK
And that the limit you're using is actually the resources (users) limit
The visible apps limit is actually housed under: `ListUsers.Filter.visibleApps`

adamjcampbell:
I'll do a quick round of testing to confirm my suspicions

Nafisa2020:
I checked the following in EndPoints-> Users-> ListUsers.swift
/// Number of resources to return.
    public enum Limit: NestableQueryParameter {
        case visibleApps(Int)

        static var key: String = "limit"
        var pair: Pair {
            switch self {
            case .visibleApps(let value):
                return ("visibleApps", "\(value)")
            }
        }
    }

adamjcampbell:
My mistake, turns out there are two top level keys for limiting things:
`"limit"` (limits the number of resources returned, in this case users)
and
`"limit[visibleApps]"` (what we're dealing with now)

So limit needs to be added to `AppStoreConnect_Swift_SDK`
But for now we should change our parameter to `--limitVisibleApps` to differentiate I think

Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
```diff
@@ -13,7 +13,7 @@ struct ListUserInvitationsCommand: CommonParsableCommand {
     @OptionGroup()
     var common: CommonOptions
 
-    @Option(help: "Limit the number of users to return (maximum 200).")
+    @Option(help: "Limit the number visible apps to return (maximum 50).")
     var limit: Int?
```
adamjcampbell, line 17: Let's change this to `limitVisibleApps`

Sources/AppStoreConnectCLI/Commands/Users/ListUsersCommand.swift
```diff
@@ -69,7 +69,7 @@ public struct ListUsersCommand: CommonParsableCommand {
             include: includeVisibleApps
                 ? [ListUsers.Include.visibleApps]
                 : nil,
-            limit: nil, // Limit of visibleApps if included, not limit of users
+            limit: limit.map { [ListUsers.Limit.visibleApps($0)] } ?? [], // Limit of visibleApps if included, not limit of users
```
adamjcampbell, line 72: I think we can remove this comment since what it is doing is described by the enum case. As well as will be described by the new variable name.

Sources/AppStoreConnectCLI/Commands/Users/Invitations/ListUserInvitationsCommand.swift
```diff
@@ -43,7 +43,7 @@ struct ListUserInvitationsCommand: CommonParsableCommand {
         let api = try makeClient()
 
         let endpoint = APIEndpoint.invitedUsers(
-            limit: limit.map { [ListInvitedUsers.Limit.visibleApps($0)] } ?? [],
+            limit: limitVisibleApps.map { [ListInvitedUsers.Limit.visibleApps($0)] } ?? [],
```
adamjcampbell, line 46: It seems `limit` takes an `Optional` array where `nil` is sent when we don't specify any limits
Therefore we can simplify this to:  

```suggestion
            limit: limitVisibleApps.map { [ListInvitedUsers.Limit.visibleApps($0)] },
```

Sources/AppStoreConnectCLI/Commands/Users/ListUsersCommand.swift
```diff
@@ -69,7 +69,7 @@ public struct ListUsersCommand: CommonParsableCommand {
             include: includeVisibleApps
                 ? [ListUsers.Include.visibleApps]
                 : nil,
-            limit: nil, // Limit of visibleApps if included, not limit of users
+            limit: limitVisibleApps.map { [ListUsers.Limit.visibleApps($0)] } ?? [],
```
adamjcampbell, line 72: Same here:

```suggestion
            limit: limitVisibleApps.map { [ListUsers.Limit.visibleApps($0)] },
```

orj:
@Adam @Nafisa2020 did this resolve #69?

Nafisa2020:
This PR fixes the limit for visible apps.
Adam's sdk change and changes to pass the user limit to the API request should fix the user limit.

PR #74, opened 2020-04-15
Correctly implement users list limit option

Nafisa2020:
# 📝 Summary of Changes

- Swift package update to include appstoreconnect-swift-sdk v 1.0.2
- Added limit-users optional argument which will limit the number of users returned

## 🔨 How To Test
swift run asc users list --limit-users 2


Sources/AppStoreConnectCLI/Commands/Users/ListUsersCommand.swift
```diff
@@ -64,12 +67,18 @@ public struct ListUsersCommand: CommonParsableCommand {
             filters.append(ListUsers.Filter.visibleApps(filterVisibleApps.compactMap { $0 }))
         }
 
+        let limits = [
+          limitUsers.map(ListUsers.Limit.users),
+          limitVisibleApps.map(ListUsers.Limit.visibleApps)
+          ]
+          .compactMap { $0 }
```
adamjcampbell, line 74: Let's align the trailing `]` with `let` and move the compact map to endpoint constructor

PR #80, opened 2020-04-17
Extract API calls out into APIOperation objects

adamjcampbell:
This is an attempt to start tackling #41 by extracting out the transformation of our defined command options into an operation to perform on the app store connect API

Feedback is very welcome
This refactor has only been performed on `ListUsersCommand` so far

# 📝 Summary of Changes

- Rename `HTTPClient` to `AppStoreConnectService`
- Create `APIOperation` to represent a an amount of work to `execute` on the API and subscribe to the result
- Create `ListUserOptions` to define what we take in to perform our operation. This `struct` is shared between the `Command` and the `Operation`
- Define the `listUsers` function on `AppStoreConnectService` that takes in our `ListUserOptions`, creates a `ListUserOperation` and executes it with it's internal `APIProvider`

# ⚠️ Items of Note

`ListUserOptions` still uses some `AppStoreConnect_Swift_SDK` defined types. We should mirror these to keep the layers separate. We may be able to easily enforce this by making a new target using the `path` option so that our commands never see the underlying SDK we are wrapping.

# 🧐🗒 Reviewer Notes

Please comment on whether this pattern makes sense or if it should be changed / modified / improved


Sources/AppStoreConnectCLI/Model/User.swift
```diff
@@ -90,7 +90,7 @@ extension User: TableInfoProvider {
     }
 }
 
-extension HTTPClient {
+extension AppStoreConnectService {
 
     /// Find the opaque internal identifier for this user; search by email adddress.
```
DechengMa, line 95:
```suggestion
    /// Find the opaque internal identifier for this user; search by email address.
```

Sources/AppStoreConnectCLI/Model/UserInvitation.swift
```diff
@@ -47,7 +47,7 @@ extension APIEndpoint where T == UserInvitationResponse {
     }
 }
 
-extension HTTPClient {
+extension AppStoreConnectService {
 
     /// Find the opaque internal identifier for this invitation; search by email adddress.
```
DechengMa, line 52:
```suggestion
    /// Find the opaque internal identifier for this invitation; search by email address.
```

Sources/AppStoreConnectCLI/Services/Options/ListUserOptions.swift
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+
+struct ListUsersOptions {
+    let limitVisibleApps: Int?
+    let limitUsers: Int?
+    let sort: ListUsers.Sort?
+    let filterUsername: [String]
```
DechengMa, line 9: I guess all of the filters can be optional when requesting?

adamjcampbell: If there are no filters we can just pass in an empty array

adamjcampbell: We could make it `[String]?` but before it was `[String?]` which wasn't very useful to us

DechengMa:
This looks good to me, if this gets merged I will apply this to all the rest of the code 💯

Sources/AppStoreConnectCLI/Services/Options/ListUserOptions.swift
adamjcampbell: Turns out there's no Option initializer for `Optional<Array<ArugumentParsable>>` so `[String]` will have to do

adamjcampbell:
I had to add more separation, namely giving each operation `Dependencies` in order to decouple it from the `APIProvider` thus making it testable.
We can't just wrap the provider in a protocol or provide a default implementation for every operation as requests rely on generics.
So we must fully qualify the type signatures for the functions that the operations use

adamjcampbell:
Also regarding testing I have seen some decent examples of helpers for testing Combine based publishers so we should look into that in a separate PR

Tests/appstoreconnect-cliTests/Operations/GetUserInfoOperationTests.swift
```diff
@@ -0,0 +1,73 @@
...
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+import XCTest
+
+typealias Dependencies = GetUserInfoOperation.Dependencies
+typealias Options = GetUserInfoOperation.Options
+typealias OperationError = GetUserInfoOperation.GetUserInfoError
+
+final class GetUserInfoOperationTests: XCTestCase {
```
orj, line 13: Tests 🎉

Sources/AppStoreConnectCLI/Services/Operations/ListUsersOperation.swift
```diff
@@ -0,0 +1,50 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+
+struct ListUsersOperation: APIOperation {
+
+    struct ListUsersDependencies {
+        let users: (APIEndpoint<UsersResponse>) -> Deferred<Future<UsersResponse, Error>>
```
orj, line 9:
```suggestion
        let users: (APIEndpoint<UsersResponse>) -> AnyPublisher<UsersResponse, Error>
```
Generally you don't want to have specific publisher types exposed.

Sources/AppStoreConnectCLI/Services/Operations/GetUserInfoOperation.swift
```diff
@@ -0,0 +1,46 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct GetUserInfoOperation: APIOperation {
+
+    struct GetUserInfoDependencies {
+        let usersResponse: (APIEndpoint<UsersResponse>) -> Deferred<Future<UsersResponse, Error>>
```
orj, line 10:
```suggestion
        let usersResponse: (APIEndpoint<UsersResponse>) -> AnyPublisher<UsersResponse, Error>
```

Sources/AppStoreConnectCLI/Services/Operations/GetUserInfoOperation.swift
```diff
@@ -0,0 +1,46 @@
...
+    init(options: GetUserInfoOptions) {
+        let filters: [ListUsers.Filter] = [.username([options.email])]
+        endpoint = APIEndpoint.users(filter: filters)
+        email = options.email
+    }
+
+    func execute(with dependencies: GetUserInfoDependencies) -> AnyPublisher<User, Error> {
+        dependencies.usersResponse(endpoint)
+            .tryMap { [email] response in
+                let users = User.fromAPIResponse(response)
```
orj, line 36:
```suggestion
                let users = [User](response)
```

Sources/AppStoreConnectCLI/Commands/Users/ListUsersCommand.swift
```diff
@@ -32,62 +32,39 @@ public struct ListUsersCommand: CommonParsableCommand {
...
+            limitVisibleApps: limitVisibleApps,
+            limitUsers: limitUsers,
+            sort: sort,
+            filterUsername: filterUsername,
+            filterRole: filterRole,
+            filterVisibleApps: filterVisibleApps,
+            includeVisibleApps: includeVisibleApps
+        )
+
+        _ = service
```
orj, line 66: What would be great would be if we could have some sort of async/await or future type construct so we didn't have to write everything as a giant combine chain.

```swift
let users = service.listUsers(with: options).await()
users.renderResult(format: common.outputFormat)
```

This I think would make writing more complex commands simpler.  It would also make the blocking explicit rather than buried in `AppStoreConnectService`.

Sources/AppStoreConnectCLI/Services/Operations/ListUsersOperation.swift
adamjcampbell: Generally no, but `Deferred` and `Future` seem to the exception to this rule.
They give us a bit of information, that the requests will only be executed on subscription (`Deferred`) and that it will only return one result (`Future`).

It occurs to me that we don't really need `Deferred` in our instances. Getting results earlier is better. So I may simplify things just to be of type `Future`.

Sources/AppStoreConnectCLI/Services/Operations/GetUserInfoOperation.swift
adamjcampbell: This change won't compile as is, would have to write an array based initializer for `User` that takes a `UserResponse`

Sources/AppStoreConnectCLI/Commands/Users/ListUsersCommand.swift
adamjcampbell: Yeah, I have been thinking about this too. It's a really good idea

adamjcampbell: I'll address this in a different PR

Tests/appstoreconnect-cliTests/Operations/GetUserInfoOperationTests.swift
adamjcampbell: 🙌

Tests/appstoreconnect-cliTests/Operations/GetUserInfoOperationTests.swift
```diff
@@ -0,0 +1,71 @@
...
+
+@testable import AppStoreConnectCLI
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+import XCTest
+
+typealias Dependencies = GetUserInfoOperation.Dependencies
+typealias Options = GetUserInfoOperation.Options
+typealias OperationError = GetUserInfoOperation.GetUserInfoError
```
DechengMa, line 11: Should we move these `typealias` into `final class GetUserInfoOperationTests`? I guess these types only need to be accessed by `GetUserInfoOperationTests`

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -0,0 +1,73 @@
...
+import Foundation
+
+class AppStoreConnectService {
+    private let provider: APIProvider
+
+    init(configuration: APIConfiguration) {
+        provider = APIProvider(configuration: configuration)
+    }
+
+    func listUsers(with options: ListUsersOptions) -> AnyPublisher<[User], Error> {
```
DechengMa, line 14: If this file getting bigger, we can probably separate these request functions into different files in the future

Tests/appstoreconnect-cliTests/Operations/GetUserInfoOperationTests.swift
```diff
@@ -0,0 +1,71 @@
...
+            )
+
+        XCTAssertEqual(
+            operationError?.localizedDescription,
+            OperationError.couldNotFindUser(email: "bob@bob.com").localizedDescription
+        )
+    }
+}
+
+private extension Dependencies {
```
DechengMa, line 44:
```suggestion
private extension GetUserInfoOperationTests.Dependencies {
```

adamjcampbell: The type alias is actually at file scope, did you want me to move the alias into the test class and extend that?

adamjcampbell: Just saw the comment above 👍

Issue #71, opened 2020-04-15
Get a beta tester's information.


orj:
# 🗣 Context

We have not yet implemented `GetBetaTesterInfoCommand`.

# 💬 Narrative

As a user
I want get all the information relating to a beta tester
So that so that I can view it.

# 📝 Notes

This should cover all the [features of the API](https://developer.apple.com/documentation/appstoreconnectapi/read_beta_tester_information).

However we should use the beta tester's email address to find them rather than their opaque resource ID as is required by the API.  So it might be easier to use the list API with a filter on email rather than call the `Read Beta Tester Information` API. I'm pretty sure the information returned is the same.

We should support `--limit-apps`, `--limit-builds`, `--limit-beta-groups` options to control the number of fetched apps, builds and betaGroups.

# 🎨 Design

```
asc testflight betatester info joe@example.com
```
Output:

| Email | First Name | Last Name | Invite Type | Apps | Beta Groups |
| :-- | :-- | :-- | :-- | :-- | :-- |
| joe@example.com | Joe | Example | EMAIL | com.iba.app1 | Friends of Friends |

In YML and JSON formats more detailed data for the App and BetaGroups could be displayed based on the data fetched we should not be restricting the field set fetched.

# ✅ Acceptance Criteria

**GIVEN** a tester exists with the email address `joe@example.com`
**WHEN** I specify the email address `joe@example.com`
**THEN** the data for the user with the email address `joe@example.com` should be displayed

**GIVEN** a tester **does not** exist with the email address `joe@example.com`
**WHEN** I specify the  email address `joe@example.com`
**THEN** an error message `Beta Tester for specified email address not found.` should be displayed on `stderr`, no output should be sent to `stdout`.

**GIVEN** a tester exists with the email address `joe@example.com` and has 10 associated apps
**WHEN** I specify the the arguments `joe@example.com --limit-apps 5`
**THEN** only 5 apps should be displayed in the list of apps the user has associated with them.

PR #82, opened 2020-04-21
TestFlight/Implement get tester information command

DechengMa:
Implement `GetBetaTesterInfoCommand`, Closes issue #71 

# 📝 Summary of Changes

- Implement `GetBetaTesterInfoCommand` according to acceptance criteria of issue #71

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage: 
` asc testflight betatesters info <email> [--limit-apps <limit-apps>] [--limit-builds <limit-builds>] [--limit-beta-groups <limit-beta-groups>]`

Example Output: 
| Email | First Name | Last Name | Invite Type | Beta Groups | Apps |
| :-- | :-- | :-- | :-- | :-- | :-- |
| joe@example.com | Joe | Example | EMAIL | Friends of Friends | com.iba.app1 |

## 🔨 How To Test
Getting an existing beta testers info
`asc testflight betatesters info youremail@gmail.com`

with limits:
`asc testflight betatesters info youremail@gmail.com --limit-beta-groups 2`


Tests/appstoreconnect-cliTests/Operations/GetBetaTestersInfoOperationTests.swift
```diff
@@ -0,0 +1,164 @@
...
+                XCTAssertEqual(expectedError, error)
+            default:
+                XCTFail("Expected failure with: \(expectedError), got: \(result)")
+        }
+    }
+}
+
+private extension GetBetaTestersInfoOperationTests.Dependencies {
+
+    static let createdSuccessResponse = """
```
DechengMa, line 56: Will change this to Fixture in a separate PR

Sources/AppStoreConnectCLI/Services/Operations/InviteTesterOperation.swift
```diff
@@ -78,12 +78,12 @@ struct InviteTesterOperation: APIOperation {
...
-            )
-            .execute(
-                with: .init(betaTesterResponse: dependencies.betaTesterResponse)
-            )
+        let endpoint = APIEndpoint.betaTester(
+            withId: testerId,
+            include: [.betaGroups, .apps])
+
+        return dependencies.betaTesterResponse(endpoint)
+            .map { BetaTester($0.data, $0.included) }
```
DechengMa, line 86: Because the requirement here is to use betatester ID to get info (more accurate), and in our `GetBetaTesterInfoOperation ` I use the email. So `GetBetaTesterInfoOperation ` can't be used here

adamjcampbell: Do you think there is room to break apart the operation for reuse?

DechengMa: We can. It will be something like `GetBetaTesterInfoByEmailOperation` and `GetBetaTesterInfoByIdOperation`. Emm... what do you think?

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/GetBetaTesterInfoCommand.swift
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct GetBetaTesterInfoCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "info",
```
adamjcampbell, line 10: @orj are we standardising all these commands as `read` commands?

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/GetBetaTesterInfoCommand.swift
```diff
@@ -0,0 +1,44 @@
...
+        let options = GetBetaTesterInfoOptions(
+            email: email,
+            limitApps: limitApps,
+            limitBuilds: limitBuilds,
+            limitBetaGroups: limitBetaGroups
+        )
+
+        let tester = try service.getBetaTestersInfo(with: options)
+            .await()
+            .first
```
adamjcampbell, line 40: How do you feel about moving `await` into the service func and having the getBetaTester (singular) function return a non optional beta tester

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -33,6 +33,10 @@ class AppStoreConnectService {
         try InviteTesterOperation(options: options).execute(with: requestor)
     }
 
+    func getBetaTestersInfo(with options: GetBetaTesterInfoOptions) throws -> AnyPublisher<[BetaTester], Error> {
+        return try GetBetaTestersInfoOperation(options: options).execute(with: requestor)
```
adamjcampbell, line 37: We can omit return now for single line returns 🎉

Sources/AppStoreConnectCLI/Services/Operations/GetBetaTestersInfoOperation.swift
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct GetBetaTestersInfoOperation: APIOperation {
+
+    private enum GetTestersError: LocalizedError {
```
adamjcampbell, line 9: Still wondering if we should just start calling these operations `Error` and have the execute signature say `Swift.Error`. May as well leave it until we decide either way

Sources/AppStoreConnectCLI/Services/Operations/GetBetaTestersInfoOperation.swift
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        return APIEndpoint.betaTesters(
+            filter: [.email([options.email])],
+            include: [.betaGroups, .apps],
+            limit: limits
+        )
+    }
+
+    let options: GetBetaTesterInfoOptions
```
adamjcampbell, line 42: You can definitely bring the Options type inside the operation and do away with the `GetBetaTesterInfoOptions` file

Sources/AppStoreConnectCLI/Services/Operations/GetBetaTestersInfoOperation.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    init(options: GetBetaTesterInfoOptions) {
+        self.options = options
+    }
+
+    func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<[BetaTester], Error> {
```
adamjcampbell, line 48: Did we want to make this return API data and transform it in the service?

Sources/AppStoreConnectCLI/Services/Operations/GetBetaTestersInfoOperation.swift
DechengMa: that will cause a pretty long function name like, eg 
```
getBetaTester(
    email: String
    limitApps: Int?
    limitBuilds: Int?
    limitBetaGroups: Int?
)
```
especially for some operations with a long list of filters 🤔 🤔 🤔 , what do you think

adamjcampbell: I'm thinking that commands don't need to really know about operations, it does result in more function params, I think that's okay. But for now if you want you can still use it like the example below while we talk about it.

E.g. `getBetaTester(options: GetBetaTesterInfoOperation.Options)`

Sources/AppStoreConnectCLI/Services/Operations/GetBetaTestersInfoOperation.swift
DechengMa: Already did this on other PRs so I will just change this to Error and Swift.Error

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -33,6 +33,29 @@ class AppStoreConnectService {
         try InviteTesterOperation(options: options).execute(with: requestor)
     }
 
+    func getBetaTesterInfo(
```
adamjcampbell, line 36: If we're returning a `BetaTester` I think we can just call it `getBetaTester`

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -33,6 +33,29 @@ class AppStoreConnectService {
...
+        email: String,
+        limitApps: Int?,
+        limitBuilds: Int?,
+        limitBetaGroups: Int?
+    ) throws -> BetaTester {
+        let operation = GetBetaTestersInfoOperation(
```
adamjcampbell, line 42: Same with the operation, I think we can call it `GetBetaTestersOperation`

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -33,6 +33,29 @@ class AppStoreConnectService {
...
+                limitBetaGroups: limitBetaGroups
+            )
+        )
+
+        let betaTesterResponse = try operation.execute(with: requestor).await()
+
+        return betaTesterResponse
+            .data
+            .map { BetaTester($0, betaTesterResponse.included) }
+            .first!
```
adamjcampbell, line 56: 🤔 I think I was confused here. If we're passing one email address into this operation we should only be getting one beta tester right?

What operation are you using for listing beta testers?
Or is that to be done in a later PR

`.first!` doesn't feel great here

Sources/AppStoreConnectCLI/Model/BetaTester.swift
```diff
@@ -29,34 +29,20 @@ struct BetaTester: ResultRenderable {
...
-            return nil
-        }
-
-        let includedBetaGroups = includes?.compactMap { relationship -> AppStoreConnect_Swift_SDK.BetaGroup? in
-            if case let .betaGroup(betaGroup) = relationship {
-                return betaGroup
-            }
-            return nil
-        }
+    init(with output: GetBetaTesterOperation.Output) {
```
adamjcampbell, line 32: For an init function that directly converts one type to another with can ommit the `with` keyword

Sources/AppStoreConnectCLI/Services/Operations/InviteTesterOperation.swift
```diff
@@ -26,7 +26,7 @@ struct InviteTesterOperation: APIOperation {
         self.options = options
     }
 
-    func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<BetaTester, Error> {
+    func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<GetBetaTesterOperation.Output, Error> {
```
adamjcampbell, line 29: This should probably just be returning the beta tester SDK model and then we can call the get operation in the service. That way we get that flat operation model and this operation is not referring to the output of another operation. What do you think?

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -30,9 +30,18 @@ class AppStoreConnectService {
...
+        let getBetaTesterOperation = GetBetaTesterOperation(
+            options: .init(
+                id: sdkBetaTester.id,
+                email: nil
+            )
+        )
+
+        let output = try getBetaTesterOperation.execute(with: requestor).await()
+
+        return BetaTester(output: output)
```
adamjcampbell: We can reduce this a lot I think:

```suggestion
        let getBetaTesterOptions = GetBetaTesterOperation.Options(id: sdkBetaTester.id, email: nil)

        return try GetBetaTesterOperation(options: getBetaTesterOptions).execute(with: requestor).await()
```

adamjcampbell: Sorry you'll need to map to an actual beta tester, but you get the idea 😄

PR #84, opened 2020-04-21
Implement awaitResult to synchronously wait for API results

adamjcampbell:
This is an attempt to move blocking calls out of `AppStoreConnectAPIService` and into a `Publisher` function called `awaitResult`.

It collects up all values and returns a `.success` with all the published values or it returns a `.failure` with the `Error` that completed the stream

# 📝 Summary of Changes

- Rename `ResultRenderable` to `Renderable`
- Rename previous `ResultRenderer` to `DefaultRenderer`
- Adds a new `ResultRenderer` that renders `Result` types
- Removes `Publisher` extensions to render results
- Adds `awaitResult` `Publisher` extension to synchronously wait for a result of all output values or the error in the stream
- Changes all `Command`s to use the new await call and result renderer

# ⚠️ Items of Note

Since the Commands currently return an `AnyPublisher` we can't know how many elements will be returned. As such `awaitResult` wraps the output values in an `Array` meaning all JSON / YAML output is surrounded in an array.

We should discuss ways to get around this. E.g. returning something like a `Future` from the `AppStoreConnectAPIService` for commands that only emit once.

Or maybe the `AppStoreConnectAPIService` functions should be blocking calls that return a result.

Alternatively we `await` for a singular result and assert that we don't get more than one

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

All existing commands should still work as expected except for the aforementioned `Array` wrapping

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/BundleIds/ModifyBundleIdCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/BundleIds/ModifyBundleIdCommand.swift
@@ -22,12 +22,14 @@ struct ModifyBundleIdCommand: CommonParsableCommand {
     var name: String
 
     func run() throws {
-        let api = try makeService()
+        let service = try makeService()
 
-        _ = try api
+        let result = try service
             .bundleIdResourceId(matching: identifier)
-            .flatMap { api.request(APIEndpoint.modifyBundleId(id: $0, name: self.name)) }
+            .flatMap { service.request(APIEndpoint.modifyBundleId(id: $0, name: self.name)) }
             .map(BundleId.init)
-            .renderResult(format: common.outputFormat)            
+            .awaitResult()
+
+        result.render(format: common.outputFormat)
     }
 }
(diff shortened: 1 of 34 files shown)


Sources/AppStoreConnectCLI/Commands/BundleIds/ModifyBundleIdCommand.swift
```diff
@@ -22,12 +22,14 @@ struct ModifyBundleIdCommand: CommonParsableCommand {
     var name: String
 
     func run() throws {
-        let api = try makeService()
+        let service = try makeService()
 
-        _ = try api
+        let result = try service
             .bundleIdResourceId(matching: identifier)
```
DechengMa, line 28: Do we want to make the most of `awaitResult` and do something like this?
```suggestion
        let bundleIdResult = try service.bundleIdResourceId(matching: identifier).awaitResult()
        let modifyResult = service.request(APIEndpoint.modifyBundleId(id: bundleIdResult, name: self.name))
            .map(BundleId.init)
            .awaitResult()
```

adamjcampbell: @DechengMa I don't think so, keep in mind `awaitResult` actually gives a swift `Result` type
If we wanted to do something like this maybe we could make it an `await` function that throws...
🤔

adamjcampbell: Going give this a go and compare...

adamjcampbell:
Closed in preference of merging #106

PR #106, opened 2020-04-22
Implement await function to synchronously wait for output throwing errors

adamjcampbell:
A throwing re-imagining of #84 

This seems like a nicer solution overall and throws `PublisherAwaitError`s if waiting doesn't go as we expect.

# 📝 Summary of Changes

- Adds `await` and `awaitMany` to synchronously wait for publisher results
- Adds `render` function to `ResultRenderable` to conveniently render it in the provided format
- Removes old `Renderers` and rendering extensions
- Errors are now thrown from a Commands `run` function

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

All existing commands should still work as expected however as we have discovered `Error`s need to be migrated to use `errorDescription` for nice printing

### DIFF
--- a/Sources/AppStoreConnectCLI/Helpers/Publisher+Helpers.swift
+++ b/Sources/AppStoreConnectCLI/Helpers/Publisher+Helpers.swift
@@ -21,8 +21,8 @@ enum PublisherAwaitError: LocalizedError {
 }
 
 extension Publisher {
-    func await(timeout: DispatchTime = .now() + 30) throws -> Output {
-        let allOutput = try awaitMany()
+    func await(timeout: DispatchTime = .distantFuture) throws -> Output {
+        let allOutput = try awaitMany(timeout: timeout)
 
         guard let output = allOutput.first, allOutput.count == 1 else {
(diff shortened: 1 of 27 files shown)


Sources/AppStoreConnectCLI/Helpers/Publisher+Helpers.swift
```diff
@@ -3,20 +3,72 @@
...
 }
 
-extension Publisher where Output == Void, Failure == Error {
-    func renderResult(format: OutputFormat) -> AnyCancellable {
-        self.sink(
-            receiveCompletion: Renderers.CompletionRenderer().render,
-            receiveValue: Renderers.null
-        )
+extension Publisher {
+    func await(timeout: DispatchTime = .now() + 30) throws -> Output {
```
DechengMa, line 24: It seems `timeout` isn't being used in this function?
```suggestion
    func await() throws -> Output {
```

adamjcampbell: Oh the timeout should be passed to the `awaitMany` function 😂 
I think I'll change the default to `distantFuture`

PR #116, opened 2020-04-22
Implement CreateCertificateCommand

### LINKED ISSUE #85: Create Certificates
# 🗣 Context

Create a new certificate using a certificate signing request.

# 💬 Narrative

As a user
I want create a certificate
So that so that I can use it with a provisioning profile.

# 📝 Notes

Covering this [API](https://developer.apple.com/documentation/appstoreconnectapi/create_a_certificate).

# 🎨 Design

```
asc certificates create ios_development --csrFile CertificateRequest.csr
```

# ✅ Acceptance Criteria

**GIVEN** I provide a valid certificate type and CSR
**WHEN** I issue a create command
**THEN** a certificate is created.

**GIVEN** I provide a invalid certificate type or CSR
**WHEN** I issue a create command
**THEN** I receive an appropriate error message.

**GIVEN** I provide valid inputs to the create command
**WHEN** the API fails
**THEN** I receive an appropriate error message.

# 🚫 Out of Scope

- Creating the CSR.  It is assumed the user has done that correctly with KeyChain Access already.

# 🤔 Future Directions

- Create the CSR for the user using macOS Security APIs.

DechengMa:
Implement create certificate command.

# 📝 Summary of Changes

- Closes #85
- Create CreateCertificateCommand
- Create struct `Certificate`
- Makes `CertificateType` `ExpressibleByArgument`

# ⚠️ Items of Note

There will also be a `certificateContent` in CertificateResponse, which is too long to render in table. So I remove it from the table but keep it when output format is `--json` or `--yaml`

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage: 
`asc certificates create <certificate-type> --csr-file <csr-file>`

Example output: 
| Name | Type | Platform | Expiration Date |
| :-- | :-- | :-- | :-- |
| Mac Installer Distribution: Jo | MAC_INSTALLER_DISTRIBUTION | MAC_OS | 2021-04-22 06:09:33 +0000 |

## 🔨 How To Test

`swift run asc certificates create IOS_DEVELOPMENT --csr-file yourFilePath/CertificateSigningRequest.certSigningRequest`

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/CreateCertificateCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/CreateCertificateCommand.swift
@@ -8,40 +8,24 @@ import Foundation
 struct CreateCertificateCommand: CommonParsableCommand {
     static var configuration = CommandConfiguration(
         commandName: "create",
-        abstract: "Create a new certificate")
+        abstract: "Create a new certificate using a certificate signing request.")
 
     @OptionGroup()
     var common: CommonOptions
 
-    @Argument(help: "The type of certificate. (eg. IOS_DEVELOPMENT)")
-    var certificateType: String
+    @Argument(help: "The type of certificate to create \(CertificateType.allCases).")
+    var certificateType: CertificateType
 
-    @Option(help: "The file path of your CSR file. (eg. folder/CertificateSigningRequest.certSigningRequest)")
+    @Argument(help: "The Certificate Signing Request (CSR) file path.")
     var csrFile: String
-
-    enum CommandError: LocalizedError {
-        case invalidCertificateType(String)
-        case invalidCSRFilePath(String)
-
-        var errorDescription: String? {
-            switch self {
-            case .invalidCertificateType(let type):
-                return "\(type) is not a valid certificate type"
-            case .invalidCSRFilePath(let path):
-                return "\(path) is not a valid CSR file path"
-            }
-        }
-    }
 
     func run() throws {
         let api = try makeService()
 
-        let type = try getCertificateType(matching: certificateType)
-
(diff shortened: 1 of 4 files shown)


orj:
@DechengMa I had a half implemented version of this command on my disk.  So I merged that code with yours.

Sources/AppStoreConnectCLI/Commands/Certificates/CreateCertificateCommand.swift
```diff
@@ -0,0 +1,37 @@
...
+struct CreateCertificateCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "create",
+        abstract: "Create a new certificate using a certificate signing request.")
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Argument(help: "The type of certificate to create \(CertificateType.allCases).")
+    var certificateType: CertificateType
```
orj, line 17: Here I'm using `CertifcateType` directly because I made it `ExpressibleByArgument`.  We use this technique elsewhere in the code.

Sources/AppStoreConnectCLI/Commands/Certificates/CreateCertificateCommand.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    @Argument(help: "The Certificate Signing Request (CSR) file path.")
+    var csrFile: String
+
+    func run() throws {
+        let api = try makeService()
+
+        let csrContent = try String(contentsOfFile: csrFile, encoding: .utf8)
```
orj, line 25: And we don't really need to translate the error `String` throws into something else.

PR #120, opened 2020-04-24
Implement Create Beta Group

### LINKED ISSUE #96: Create a Beta Group
# 🗣 Context

Create a beta group associated with an app, optionally enabling TestFlight public links.

# 💬 Narrative

As a user
I want create beta groups for my app
So that I can start adding beta testers to them

# 📝 Notes

Covering this [API](https://developer.apple.com/documentation/appstoreconnectapi/create_a_beta_group).

# 🎨 Design

```
asc testflight betagroup create appBundleId groupName --[no]-public-link --public-link-limit <n> 
```

**Notes:**
- We can use `@Flag`'s support for negation with `no` prefix.
- The specification of a `--public-link-limit` implies that `publicLinkLimitEnabled` should be true in the `BetaGroupCreateRequest.Data.Attributes`.

**Output:**
| App Bundle ID | App Name | Group Name | Is Internal | Public Link | Public Link Enabled | Public Link Limit | Public Link Limit Enabled | Creation Date |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |

# ✅ Acceptance Criteria

**GIVEN** I provide invalid input
**WHEN** I issue the command
**THEN** an appropriate error message should be displayed

**GIVEN** I provide valid input
**WHEN** I issue the command
**THEN** the output should be displayed

adamjcampbell:
This PR implements the ability to create beta groups
Closes #96 

# 📝 Summary of Changes

- Adds `BetaGroupCommand`, `CreateBetaGroupCommand`, `BetaGroup`, `GetAppsOperation`, `CreateBetaGroupOperation` and `CreateBetaGroupOptions`

# ⚠️ Items of Note

- I'm not sure the approach of testing the Endpoint body is the best, feel free to suggest alternatives
- I could not get the public link options to work, I think it requires builds to have been uploaded prior. Requires more testing.

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage: `asc betagroup create [--api-issuer <uuid>] [--api-key-id <keyid>] [--csv] [--json] [--table] [--yaml] <app-bundle-id> <group-name> [--public-link] [--no-public-link] [--public-link-limit <public-link-limit>]`

## 🔨 How To Test

Command: 
`swift run asc betagroup create com.test1 testGroup`
Sample Output:
```
+---------------+--------------+------------+-------------+-------------+---------------------+-------------------+---------------------------+---------------------------+
| App Bundle ID | App Name     | Group Name | Is Internal | Public Link | Public Link Enabled | Public Link Limit | Public Link Limit Enabled | Creation Date             |
+---------------+--------------+------------+-------------+-------------+---------------------+-------------------+---------------------------+---------------------------+
| com.test1     | ABC Test App | testGroup  | false       |             | false               |                   | false                     | 2020-04-24 07:12:07 +0000 |
+---------------+--------------+------------+-------------+-------------+---------------------+-------------------+---------------------------+---------------------------+
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/Operations/GetAppsOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/GetAppsOperation.swift
@@ -4,8 +4,8 @@ import AppStoreConnect_Swift_SDK
 import Combine
 import Foundation
 
-struct GetAppIdsOperation: APIOperation {
-    struct GetAppIdsDependencies {
+struct GetAppsOperation: APIOperation {
+    struct GetAppsDependencies {
         let apps: (APIEndpoint<AppsResponse>) -> Future<AppsResponse, Error>
     }
 
@@ -23,18 +23,18 @@ struct GetAppIdsOperation: APIOperation {
         }
     }
 
-    private let options: GetAppIdsOptions
+    private let options: GetAppsOptions
 
-    init(options: GetAppIdsOptions) {
+    init(options: GetAppsOptions) {
         self.options = options
     }
 
-    func execute(with dependencies: GetAppIdsDependencies) -> AnyPublisher<[String], Error> {
+    func execute(with dependencies: GetAppsDependencies) -> AnyPublisher<[AppStoreConnect_Swift_SDK.App], Error> {
         let bundleIds = options.bundleIds
         let endpoint = APIEndpoint.apps(filters: [.bundleId(bundleIds)])
 
         return dependencies.apps(endpoint)
-            .tryMap { (response: AppsResponse) throws -> [String] in
+            .tryMap { (response: AppsResponse) throws -> [AppStoreConnect_Swift_SDK.App] in
                 guard !response.data.isEmpty else {
                     throw GetAppIdsError.couldntFindAnyAppsMatching(bundleIds: bundleIds)
                 }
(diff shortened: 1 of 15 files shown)


Sources/AppStoreConnectCLI/Services/Operations/GetAppsOperation.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    private let options: GetAppsOptions
+
+    init(options: GetAppsOptions) {
+        self.options = options
+    }
+
+    func execute(
+        with dependencies: GetAppsDependencies
+    ) -> AnyPublisher<[AppStoreConnect_Swift_SDK.App], Error> {
```
DechengMa, line 34: We have our own App Model as well, and I understand sometimes we need the id from the SDK model to fetch further information. I was wondering how do we solve the incompatible? eg. Can we use our `App` model here: `AnyPublisher<[Apps], Error>`?

adamjcampbell: I think we can move to our own App model, yep

DechengMa: > I think we can move to our own App model, yep

But I think you need the APP internal resource id when fetching stuff. Our App Model doesn't have that. What are your thoughts on this?

adamjcampbell: I think that gives us more of a reason to separate our models and networking into a separate module. That way we can have the internal id be `internal` access level and the rest of it be `public`

PR #121, opened 2020-04-27
Add ListPreReleaseVersionsCommand

### LINKED ISSUE #102: List Prerelease Versions
# 🗣 Context

Get a list of prerelease versions for all apps.

# 💬 Narrative

As a user
I want to list pre-release versions of my application
So that I can inspect the details

# 📝 Notes

Covering this [API](https://developer.apple.com/documentation/appstoreconnectapi/list_prerelease_versions).

# 🎨 Design

```
asc testflight prereleaseversion list --filter-app <bundleId> --filter-build-processing-state <states...> --filter-builds-expired <bool?> --filter-platform <platforms...> --filter-version <versions...> --limit <n> --sort <order>
```

# ✅ Acceptance Criteria

**GIVEN** ...
**WHEN** ...
**THEN** ...

# 🚫 Out of Scope

_Describe what is out of scope_

orj:
# 📝 Summary of Changes

- Resolves #102

# ⚠️ Items of Note

- Introduces a `ListPreReleaseVersionCommand` command.
- Adds coverage gathering on unit tests
- Depends on a fix to the AppStoreConnect SDK

# 🧐🗒 Reviewer Notes

I'm interested in feedback on the tests and the style of `ListPreReleaseVersionsOperation`.

## 💁 Example

```
$ swift run asc testflight prereleaseversion list
+------------+---------------------------------+-----------------------+----------+---------+
| App ID     | App Bundle ID                   | App Name              | Platform | Version |
+------------+---------------------------------+-----------------------+----------+---------+
| 1234567890 | com.example.ApplicationName     | Example App           | ios      | 1.0     |
+------------+---------------------------------+-----------------------+----------+---------+
```

When you specify a bundle ID that doesn't exist you get no results instead of an error.
```
$ swift run asc testflight prereleaseversion list --filter-bundle-ids xcom.example
+--------+---------------+----------+----------+---------+
| App ID | App Bundle ID | App Name | Platform | Version |
+--------+---------------+----------+----------+---------+

+--------+---------------+----------+----------+---------+
```

## 🔨 How To Test

Run the command to list pre-release versions.  Try some of the filtering options.


README.md
```diff
@@ -116,9 +116,33 @@ This project is licensed under the terms of the MIT license. See the [LICENSE][l
 
 > This project is in no way affiliated with [Apple Inc][apple]. This project is open source under the MIT license, which means you have full access to the source code and can modify it to fit your own needs. _AppStoreConnect CLI_ runs on your own computer or server and does not communicate with any service other than the [Apple AppStore Connect API][appstore-connect-api]. You are responsible for how you use _AppStoreConnect CLI_.
 
+## Acknowledgements
+
+We use the following Swift Packages in `appstoreconnect-cli`.  We are very greatful to their authors.
```
adamjcampbell, line 121:
```suggestion
We use the following Swift Packages in `appstoreconnect-cli`.  We are very grateful to their authors.
```

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @OptionGroup()
+    var options: Options
+
+    func validate() throws {
+        if options.filterAppId.isEmpty == false && options.filterBundleId.isEmpty == false {
+            throw ValidationError("Filtering by both Bundle ID and App ID is not supported!")
```
adamjcampbell, line 69: Good to know this `validate` func is here: I specified that appIds supercede bundleIds in my command but this is probably a better approach

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,92 @@
...
+import AppStoreConnect_Swift_SDK
+import ArgumentParser
+import Foundation
+
+struct ListPreReleaseVersionsCommand: CommonParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "list",
+        abstract: "Get a list of prerelease versions for all apps.")
+
+    struct Options: ParsableArguments {
```
adamjcampbell, line 12: 🤔 I'm seeing some commonality between some of these arguments and some arguments in other commands

I think we should maybe chunk option groups together to share them between commands, especially to share the text

I know you were trying to use this group to translate into the operation options. But I think they can have different purposes

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    func validate() throws {
+        if options.filterAppId.isEmpty == false && options.filterBundleId.isEmpty == false {
+            throw ValidationError("Filtering by both Bundle ID and App ID is not supported!")
+        }
+    }
+
+    func run() throws {
+        try makeService()
+            .listPreReleaseVersions(with: .init(options))
```
adamjcampbell, line 75: I had a go at just passing through function params to the service in order to keep operations kind of internal to how to service works.
Do you think that's worthwhile?

I also used `await` inside making the function service throwing so that the service directly returned our models

adamjcampbell:
I think we now have some differing approaches between this and #123 for which we might want to meet somewhere in the middle

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import ArgumentParser
+import Foundation
```
DechengMa, line 5:
```suggestion
```

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,73 @@
...
+    @Option(
+        parsing: .unconditional,
+        help: ArgumentHelp(
+            "Sort the results using the provided key \(ListPrereleaseVersions.Sort.allCases).",
+            discussion: "The `-` prefix indicates descending order."
+        )
+    )
+    var sort: ListPrereleaseVersions.Sort?
+    
+
```
DechengMa, line 62:
```suggestion
```

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,73 @@
...
+    @Option(
+        parsing: .upToNextOption,
+        help: ArgumentHelp(
+            "Filter by app AppStore ID. eg. 432156789",
+            discussion: "This option is mutually exclusive with --filter-bundle-id.",
+            valueName: "app-id"
+        )
+    )
+    var filterAppIds: [String]
```
DechengMa, line 33: In ReadAppCommand we have something like this, can you try to reuse it? (eg. may be move this to .AppStoreConnectCLI/Arguments) and use an `@OptionGroup`
```suggestion
@Option(
        help: ArgumentHelp(
            "The app AppStore ID. eg. 432156789 or app bundle identifier. eg. com.example.App",
            discussion: "Please input either app id or bundle Id",
            valueName: "app-id / bundle-id"
        ),
        transform: Identifier.init
    ) var identifier: Identifier

    enum Identifier {
        case appId(String)
        case bundleId(String)

        init(_ argument: String) {
            switch Int(argument) == nil {
            case true:
                self = .bundleId(argument)
            case false:
                self = .appId(argument)
            }
        }
    }
```

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
```diff
@@ -0,0 +1,73 @@
...
+    )
+    var sort: ListPrereleaseVersions.Sort?
+    
+
+    func run() throws {
+        let service = try makeService()
+
+        let prereleaseVersions = filterAppIds.isEmpty
+            ? try service.listPreReleaseVersions(filterBundleIds: filterBundleIds, filterVersions: filterVersions, filterPlatforms: filterPlatforms, sort: sort)
+            : try service.listPreReleaseVersions(filterAppIds: filterAppIds, filterVersions: filterVersions, filterPlatforms: filterPlatforms, sort: sort)
```
DechengMa, line 68: An enum can be used here if the above changes were performed
```suggestion
        let prereleaseVersions = try service.listPreReleaseVersions(identifier: identifier, filterVersions: filterVersions, filterPlatforms: filterPlatforms, sort: sort)
```

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
```diff
@@ -0,0 +1,57 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Foundation
+import SwiftyTextTable
+
+struct PreReleaseVersionDetails: ResultRenderable {
```
DechengMa, line 7: Can have a discussion if we want to keep the consistency or not, we haven't used keyword `Details` for other models.
```suggestion
struct PreReleaseVersion: ResultRenderable {
```

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
```diff
@@ -0,0 +1,57 @@
...
+        let relationships = preReleaseVersion.relationships
+
+        let includedApps = includes?.compactMap { relationship -> AppStoreConnect_Swift_SDK.App? in
+          if case let .app(app) = relationship {
+            return app
+          }
+          return nil
+        }
+
+        let appDetails = includedApps?.filter { relationships?.app?.data?.id == $0.id }.first
```
DechengMa, line 26:
```suggestion
        let appDetails = includedApps?.first { relationships?.app?.data?.id == $0.id }
```

Sources/AppStoreConnectCLI/Model/ProcessingState.swift
```diff
@@ -0,0 +1,15 @@
...
+enum ProcessingState: String, CaseIterable, CustomStringConvertible {
+    case processing = "PROCESSING"
+    case failed = "FAILED"
+    case invalid = "INVALID"
+    case valid = "VALID"
+
+    public var description: String {
+        rawValue.lowercased()
+    }
+}
```
DechengMa, line 15: Is this file being used?

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -428,6 +428,52 @@ class AppStoreConnectService {
 
         return App(sdkApp)
     }
+
+    func listPreReleaseVersions(
+        filterAppIds: [String],
```
DechengMa, line 433: See above
```suggestion
        identifier: App.Identifier,
```

Sources/AppStoreConnectCLI/Services/Operations/ListPreReleaseVersionsOperation.swift
```diff
@@ -0,0 +1,45 @@
...
+        var filterVersions: [String] = []
+        var filterPlatforms: [String] = []
+        var sort: ListPrereleaseVersions.Sort? = nil
+    }
+
+    typealias PreReleaseVersion =  AppStoreConnect_Swift_SDK.PrereleaseVersion
+    typealias Relationships = [AppStoreConnect_Swift_SDK.PreReleaseVersionRelationship]?
+    typealias Output = [(preReleaseVersion: PreReleaseVersion, relationships: Relationships)]
+
+
```
DechengMa, line 20:
```suggestion
```

Sources/AppStoreConnectCLI/Services/Operations/ListPreReleaseVersionsOperation.swift
```diff
@@ -0,0 +1,45 @@
...
+        let endpoint = APIEndpoint.prereleaseVersions(
+            filter: filters,
+            include: [.app],
+            sort: [options.sort].compactMap { $0 }
+        )
+
+        return requestor.request(endpoint)
+            .map{ response -> Output in
+                return response.data.map {($0, response.included)}
```
DechengMa, line 41:
```suggestion
                return response.data.map { ($0, response.included) }
```

Sources/AppStoreConnectCLI/Model/ProcessingState.swift
Nafisa2020: I didn't add this..it was added by whoever started this PR

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
Nafisa2020: I don't think this is correct. You should first filter by condition and then take the first occurrence.

DechengMa: I think `array.filter` will iterate the array and `array.first` will find the first then stop looking the others. I guess `array.first` will have a better performance? @adamjcampbell What's your thought on this?

Nafisa2020: I think filtering first is more safer because if you are only considering the first occurrence and the one that you want is not at the first occurrence then you are missing out the one you actually want.

adamjcampbell: @Nafisa2020 this is actually the function [`first(where:)`](https://developer.apple.com/documentation/swift/array/1848165-first)
It will iterate and stop when it finds the first thing that matches, it's a very useful function

Nafisa2020: then it makes sense otherwise blindly assuming it will be always at the first occurrence is not good.

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
adamjcampbell: Creating an option group for this seems like a good idea since we use it quite a bit

Nafisa2020: I need to use the identifier as an option in list command and as an argument in read command. So will one option group work?

adamjcampbell: Might need two in that case, but we are using them for other commands so we should still get some reuse

Nafisa2020: well, list command should be able to take an array of identifiers..so we may need three

Nafisa2020: Maybe we should have a separate PR for refactoring this..at the moment maybe I should keep it in individual files..what's your thought on this?

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
DechengMa: Yeh it's actually the func `first(where: )`, sometimes omit the keyword `where` is not a good idea 🤔

Sources/AppStoreConnectCLI/Commands/TestFlight/PreReleaseVersion/ListPreReleaseVersionsCommand.swift
Nafisa2020: Passing an array of identifier.

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
Nafisa2020: Passed as func listPreReleaseVersions(
        filterIdentifiers: [ListPreReleaseVersionsCommand.Identifier],

PR #122, opened 2020-04-27
Implement Read download certificate Command

### LINKED ISSUE #88: Read & Download Certificate
# 🗣 Context

Get information about a certificate and download the certificate data.

# 💬 Narrative

As a user
I want to display information about a certificate and optionally download the certificate file
So that I can inspect the certificate

# 📝 Notes

Covering this [API](https://developer.apple.com/documentation/appstoreconnectapi/read_and_download_certificate_information).

# 🎨 Design

```
asc certificates read  XXYYZZ --certificate-output ./file.cer 
```

Output:

"📥 Certificate '\(serial)' downloaded to: \(certificateOutput)"

SerialNumber | Name | Display Name | Platform | Type | Expiration Date
-- | -- | -- | -- | -- | --
XXXYY | Name | Something | iOS | ios_development | Date

# ✅ Acceptance Criteria

GIVEN I provide a valid input without `--certificate-output`
WHEN I issue a read command
THEN the certificate's information is displayed (as above).

GIVEN I provide valid input with `--certificate-output`
WHEN I issue a list command
THEN the certificate's information is displayed and the certificate data is written to the specified file.

GIVEN I provide a invalid input
WHEN I issue a list command
THEN I receive an appropriate error message.

GIVEN I provide valid inputs to the list command
WHEN the API fails
THEN I receive an appropriate error message.

# 🚫 Out of Scope

- Displaying the decoded certificate contents.

# 🤔 Future Directions

- The the X.509 certificate information could be displayed like:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            10:e6:fc:62:b7:41:8a:d5:00:5e:45:b6
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C=BE, O=GlobalSign nv-sa, CN=GlobalSign Organization Validation CA - SHA256 - G2
        Validity
            Not Before: Nov 21 08:00:00 2016 GMT
            Not After : Nov 22 07:59:59 2017 GMT
        Subject: C=US, ST=California, L=San Francisco, O=Wikimedia Foundation, Inc., CN=*.wikipedia.org
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
            pub: 
                    00:c9:22:69:31:8a:d6:6c:ea:da:c3:7f:2c:ac:a5:
                    af:c0:02:ea:81:cb:65:b9:fd:0c:6d:46:5b:c9:1e:
                    9d:3b:ef
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Agreement
            Authority Information Access: 
                CA Issuers - URI:http://secure.globalsign.com/cacert/gsorganizationvalsha2g2r1.crt
                OCSP - URI:http://ocsp2.globalsign.com/gsorganizationvalsha2g2
            X509v3 Certificate Policies: 
                Policy: 1.3.6.1.4.1.4146.1.20
                  CPS: https://www.globalsign.com/repository/
                Policy: 2.23.140.1.2.2
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 CRL Distribution Points: 
                Full Name:
                  URI:http://crl.globalsign.com/gs/gsorganizationvalsha2g2.crl
            X509v3 Subject Alternative Name: 
                DNS:*.wikipedia.org, DNS:*.m.mediawiki.org, DNS:*.m.wikibooks.org, DNS:*.m.wikidata.org, DNS:*.m.wikimedia.org, DNS:*.m.wikimediafoundation.org, DNS:*.m.wikinews.org, DNS:*.m.wikipedia.org, DNS:*.m.wikiquote.org, DNS:*.m.wikisource.org, DNS:*.m.wikiversity.org, DNS:*.m.wikivoyage.org, DNS:*.m.wiktionary.org, DNS:*.mediawiki.org, DNS:*.planet.wikimedia.org, DNS:*.wikibooks.org, DNS:*.wikidata.org, DNS:*.wikimedia.org, DNS:*.wikimediafoundation.org, DNS:*.wikinews.org, DNS:*.wikiquote.org, DNS:*.wikisource.org, DNS:*.wikiversity.org, DNS:*.wikivoyage.org, DNS:*.wiktionary.org, DNS:*.wmfusercontent.org, DNS:*.zero.wikipedia.org, DNS:mediawiki.org, DNS:w.wiki, DNS:wikibooks.org, DNS:wikidata.org, DNS:wikimedia.org, DNS:wikimediafoundation.org, DNS:wikinews.org, DNS:wikiquote.org, DNS:wikisource.org, DNS:wikiversity.org, DNS:wikivoyage.org, DNS:wiktionary.org, DNS:wmfusercontent.org, DNS:wikipedia.org
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Subject Key Identifier: 
                28:2A:26:2A:57:8B:3B:CE:B4:D6:AB:54:EF:D7:38:21:2C:49:5C:36
            X509v3 Authority Key Identifier: 
                keyid:96:DE:61:F1:BD:1C:16:29:53:1C:C0:CC:7D:3B:83:00:40:E6:1A:7C

    Signature Algorithm: sha256WithRSAEncryption
         8b:c3:ed:d1:9d:39:6f:af:40:72:bd:1e:18:5e:30:54:23:35:
         ...
```

DechengMa:
Implement Read and download certificate command, closes #88 

# 📝 Summary of Changes

- Create ReadCertificateOption and Operation
- Implement ReadDownload certificate
- Create Tests for file extension helper function
- Create Tests for ReadDownloadCetificate Operation.
- Separate Mockup data for reuse purpose

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage: 

`asc certificates read <serial> [--certificate-output <certificate-output>]`

Sample output:
`📥 Certificate 'Mac Installer Distribution: Foo' downloaded to: /Documents/github/foo.cer`

SerialNumber | Name | Display Name | Platform | Type | Expiration Date
-- | -- | -- | -- | -- | --
XXXYY | Mac Installer Distribution: Foo | Something | MAC_INSTALLER_DISTRIBUTION | MAC_OS | 2021-04-22 08:02:15 +0000

## 🔨 How To Test

` swift run asc certificates read ABCD1234 --certificate-output ../foo.cer`


Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
```diff
@@ -70,19 +56,19 @@ struct ListDownloadCertificates: CommonParsableCommand {
         if let downloadPath = downloadPath {
             try certificates.forEach { (certificate: Certificate) in
                 guard let content = certificate.content else {
-                    throw CommandError.invalidContent
+                    throw CertificatesError.invalidContent
```
orj, line 59: 🤔 This error might be better named `noContent`.
```suggestion
                    throw CertificatesError.noContent
```

Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
```diff
@@ -70,19 +56,19 @@ struct ListDownloadCertificates: CommonParsableCommand {
...
                 }
 
                 do {
-                    let folder = try Folder(path: downloadPath)
-
-                    let file = try folder.createFile(
+                    let file = try File.createFile(
```
orj, line 63: 🤔 What sort of error does this throw? Could we just use that error rather than mapping to `CertificatesError.invalidPath`?

Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
```diff
@@ -6,7 +6,7 @@ import Combine
 import Foundation
 import Files
 
-struct ListDownloadCertificates: CommonParsableCommand {
+struct ListDownloadCertificatesCommand: CommonParsableCommand {
```
orj, line 9: 👍

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,57 @@
...
+
+                print("📥 Certificate '\(certificate.name ?? "")' downloaded to: \(file.path)")
+            } catch {
+                throw CertificatesError.invalidPath(certificateOutput)
+            }
+        }
+
+        certificate.render(format: common.outputFormat)
+    }
+
```
orj, line 56:
```suggestion
```

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,57 @@
...
+        if let certificateOutput = certificateOutput {
+            guard let content = certificate.content else {
+                throw CertificatesError.invalidContent
+            }
+
+            do {
+                let (folderName, fileName) = File.getFolderAndFileName(from: certificateOutput)
+
+                let file = try File.createFile(
```
orj, line 42: 🤔 Again I wonder if remapping this thrown error is actually worth while.

Sources/AppStoreConnectCLI/Helpers/File+Helper.swift
```diff
@@ -0,0 +1,39 @@
...
+    ) throws -> File {
+        let folder = try Folder(path: folderName)
+
+        return try folder.createFile(
+            named: fileName,
+            contents: Data(base64Encoded: content)
+        )
+    }
+
+    static func getFolderAndFileName(from path: String) -> (String, String) {
```
orj, line 20:
```suggestion
    static func folderAndFilename(from path: String) -> (String, String) {
```

Though most of this function is unnecessary. You should probably just be using `URL` apis.

Eg:
```
        let f = URL(fileURLWithPath: "test/test")
        f.deletingLastPathComponent().path
        f.lastPathComponent
```

Though URL has issues with relative paths in the format of "./test".  You get `/` as the folder which is wrong.

So you can use `NSString` APIs:

```
let f: NSString = "./test"
let f2 = f.standardizingPath as NSString

f2.lastPathComponent
f2.deletingLastPathComponent
```
`standardizingPath` will remove any `../.././` stuff from the path.

So if `deletingLastPathComponent` returns an empty string then you know that the path is relative.

Unfortunately Apple hasn't put any of these path manipulation APIs on `String` in Swift.  I guess they consider them deprecated or not ideal so aren't exposing them on `String`.

They've keep the URL APIs.  But relative paths and file:// URLs are problematic.

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+        serial = options.serial
+    }
+
+    func execute(with dependencies: ReadCertificateDependencies) -> AnyPublisher<Certificate, Error> {
+        dependencies
+            .certificatesResponse(endpoint)
+            .tryMap { [serial] (response: CertificatesResponse) -> Certificate in
+                switch response.data.count {
```
orj, line 43: 👍

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    enum ReadCertificateError: LocalizedError {
+        case couldNotFindCertificate(String)
+        case serialNumberNotUnique(String)
+
+        var errorDescription: String? {
+            switch self {
+            case .couldNotFindCertificate(let serial):
+                return "Couldn't find certificate with input '\(serial)'"
```
orj, line 20:
```suggestion
                return "Couldn't find certificate with serial '\(serial)'."
```

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+            case .serialNumberNotUnique(let serial):
+                return "The serial number your input '\(serial)' is not unique"
```
orj, line 22:
```suggestion
                return "The serial number provided '\(serial)' is not unique."
```

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    private let serial: String
+
+    init(options: ReadCertificateOptions) {
+        endpoint = APIEndpoint.listDownloadCertificates(
+            filter: [.serialNumber([options.serial])]
+        )
+
+        serial = options.serial
+    }
```
orj, line 37:
```suggestion
    private var endpoint: APIEndpoint<CertificatesResponse> {
        APIEndpoint.listDownloadCertificates(
            filter: [.serialNumber([options.serial])]
        )
    }

    let options: ReadCertificateOptions    
```

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    func execute(with dependencies: ReadCertificateDependencies) -> AnyPublisher<Certificate, Error> {
+        dependencies
+            .certificatesResponse(endpoint)
+            .tryMap { [serial] (response: CertificatesResponse) -> Certificate in
```
orj, line 42:
```suggestion
            .tryMap { [serial = options.serial] (response: CertificatesResponse) -> Certificate in
```

Sources/AppStoreConnectCLI/Model/Errors/Certificates.swift
```diff
@@ -0,0 +1,17 @@
...
+import Foundation
+
+enum CertificatesError: LocalizedError {
+    case invalidPath(String)
+    case invalidContent
+
+    var errorDescription: String? {
+        switch self {
+        case .invalidPath(let path):
+            return "Download failed, please check the path '\(path)' you input and try again"
```
orj, line 12:
```suggestion
            return "Download failed, please check the path '\(path)' you input and try again."
```

Sources/AppStoreConnectCLI/Model/Errors/Certificates.swift
```diff
@@ -0,0 +1,17 @@
...
+        case .invalidContent:
+            return "The certificate in the response doesn't have a proper content"
```
orj, line 14:
```suggestion
            return "The certificate in the response doesn't have a proper content."
```

Sources/AppStoreConnectCLI/Helpers/File+Helper.swift
```diff
@@ -0,0 +1,39 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Files
+
+extension File {
+    static func createFile(
+        in folderName: String,
+        named fileName: String,
+        with content: String
```
orj, line 10: It is not obvious that `content` is expected to be a Base64 encoded string.

Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
```diff
@@ -70,19 +56,19 @@ struct ListDownloadCertificates: CommonParsableCommand {
...
 
                 do {
-                    let folder = try Folder(path: downloadPath)
-
-                    let file = try folder.createFile(
+                    let file = try File.createFile(
+                        in: downloadPath,
                         named: "\(certificate.serialNumber ?? "serial").cer",
-                        contents: Data(base64Encoded: content)
+                        with: content
```
orj, line 66: I'm not sure this is actually necessary to be extracted out to a helper method.
This code is as simple when it is just:
```swift
try Folder(path: downloadPath)
    .createFile(
        named:"\(certificate.serialNumber ?? "serial").cer", 
        contents: Data(base64Encoded: content)
    )
```

Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
```diff
@@ -69,21 +55,18 @@ struct ListDownloadCertificates: CommonParsableCommand {
 
         if let downloadPath = downloadPath {
             try certificates.forEach { (certificate: Certificate) in
-                guard let content = certificate.content else {
-                    throw CommandError.invalidContent
+                guard let content = certificate.content,
+                    let data = Data(base64Encoded: content) else {
+                    throw CertificatesError.noContent
```
adamjcampbell, line 60: Don't know about the rest of the team but I can find the following formatting confusing as the indentation for the `guard let` bindings is the same for the `throw`

```suggestion
                guard 
                    let content = certificate.content,
                    let data = Data(base64Encoded: content)
                else {
                    throw CertificatesError.noContent
```

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    @Option(help: "The file download path and name. (eg. ./file.cer)")
+    var certificateOutput: String?
+
+    func run() throws {
+        let service = try makeService()
+
+        let certificate = try service
+            .readCertificate(serial: serial)
+            .await()
```
adamjcampbell, line 28: We could choose to put await in the service function here

Sources/AppStoreConnectCLI/Helpers/File+Helper.swift
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Files
+
+extension File {
+    static func folderAndFilename(from path: String) -> (String, String) {
+        let filePath: NSString = path as NSString
+        
+        return (filePath.deletingLastPathComponent, filePath.lastPathComponent)
```
adamjcampbell, line 10: You can name your tuple arguments here for increased readability

```suggestion
    static func folderAndFilename(from path: String) -> (folderName: String, fileName: String) {
        let filePath: NSString = path as NSString
        
        return (folderName: filePath.deletingLastPathComponent, fileName: filePath.lastPathComponent)
```

Sources/AppStoreConnectCLI/Model/Errors/Certificates.swift
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+enum CertificatesError: LocalizedError {
```
adamjcampbell, line 5: 🤔 we should standardise where we put errors

Sources/AppStoreConnectCLI/Commands/Certificates/ListDownloadCertificatesCommand.swift
orj: I've done this in the past. I think it helps with readability too.

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
orj: Is the service intended to be blocking @adamjcampbell? I can see reasons for both sides of this argument. Ie, service could be async vs entirely blocking (for simplicity).

Sources/AppStoreConnectCLI/Helpers/File+Helper.swift
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Files
+
+extension File {
+    static func folderAndFilename(from path: String) -> (String, String) {
```
orj, line 7: I don't think we actually need this method. It is also an extension on `File` which isn't accurate.

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        if let certificateOutput = certificateOutput {
+            guard let content = certificate.content,
+                let data = Data(base64Encoded: content) else {
+                throw CertificatesError.noContent
+            }
+
+            let (folderName, fileName) = File.folderAndFilename(from: certificateOutput)
```
orj, line 36:
```suggestion
            let standardizedPath = certificateOutput.standardizedPath as NSString
            let folderName = standardizedPath.deletingLastPathComponent
            let folderName = standardizedPath.lastPathComponent
```

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,49 @@
...
+
+            let file = try Folder(path: folderName)
```
orj, line 38:
```suggestion
            let file = try Folder(path: standardizedPath.deletingLastPathComponent)
```

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
```diff
@@ -0,0 +1,49 @@
...
+                .createFile(
+                    named: fileName,
```
orj, line 40:
```suggestion
                    named: standardizedPath.lastPathComponent,
```

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
orj: My only concern with the suggestion is to understand what `Folder(path:)` does when you pass an empty string.

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
```diff
@@ -0,0 +1,54 @@
...
+import Combine
+import Foundation
+
+struct ReadCertificateOperation: APIOperation {
+
+    struct Options {
+        let serial: String
+    }
+
+    enum ReadCertificateError: LocalizedError {
```
orj, line 13: If you're nesting a type like this then you don't need a type like `ReadCertificateError` it could just be `Error`.

orj: This is because its actual type name is `ReadCertificateOperation.ReadCertificateError` which seems a bit repetitive when it could be `ReadCertificateOperation.Error`.

orj: Admittedly this will introduce a disambiguation issue with `Swift.Error`. You would have to make sure the methods below like `execute ` disambiguated `Error`.

Tests/appstoreconnect-cliTests/Operations/ReadCertificateOperationTests.swift
```diff
@@ -0,0 +1,117 @@
...
+        case .failure(let error as OperationError):
+            XCTAssertEqual(error.errorDescription, expectedError.errorDescription)
+        default:
+            XCTFail("Expected failed with \(expectedError), got: \(result)")
+        }
+    }
+
+    static let readCertificateResponse = Certificate
+        .readCertificateResponse
+        .map({ try! jsonDecoder.decode(CertificatesResponse.self, from: $0) })!
```
orj, line 107: FYI: There are problems with things like `try!` in tests.  Admittedly these uses should be OK because you're in control of the JSON and it is a programmer error for the JSON to be invalid.  But generally you should throw in tests  and make your test methods be marked as `throws`.

Sources/AppStoreConnectCLI/Services/Operations/ReadCertificateOperation.swift
adamjcampbell: Luckily we are slightly protected because the protocol defines `Swift.Error` so you are forced to disambiguate

Tests/appstoreconnect-cliTests/Operations/ReadCertificateOperationTests.swift
adamjcampbell: How have you found throwing tests? I have found the output to be sub optimal when they do throw.

Sources/AppStoreConnectCLI/Commands/Certificates/ReadDownloadCertificateCommand.swift
adamjcampbell: I've come down on the side of blocking for simplicity inside the service we can decide what is parallel and what is not

adamjcampbell:
🤔 since the command names are `read` and `list` we should probably simplify the names to `ReadCertificateCommand` and `ListCertificatesCommand` respectively

PR #132, opened 2020-05-05
Test modify beta group operation

adamjcampbell:
Adds some unit tests for modify beta group

# 📝 Summary of Changes

- Test success
- Test that the correct errors are thrown
- Test that the request body is formed with the passed options

### DIFF
--- a/Tests/appstoreconnect-cliTests/Operations/ModifyBetaGroupOperationTests.swift
+++ b/Tests/appstoreconnect-cliTests/Operations/ModifyBetaGroupOperationTests.swift
@@ -32,10 +32,10 @@ final class ModifyBetaGroupOperationTests: XCTestCase {
         Options(
             app: app,
             currentGroupName: "Some Group",
-            newGroupName: nil,
-            publicLinkEnabled: nil,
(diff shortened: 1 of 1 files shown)


Tests/appstoreconnect-cliTests/Operations/ModifyBetaGroupOperationTests.swift
```diff
@@ -0,0 +1,166 @@
...
+            included: nil,
+            links: PagedDocumentLinks(first: nil, next: nil, self: self.testUrl),
+            meta: nil
+        )
+
+        return Future { $0(.success(response)) }
+    }
+
+    func testSuccess() throws {
+        let successRequestor = TwoEndpointTestRequestor(
```
DechengMa, line 73: `TwoEndpointTestRequestor ` is a nice one! but what if there are more than two requests in an Operation 🤔

adamjcampbell: Then we'll need to add another requestor type... It's not my favorite solution. But I think it's what we have unless the SDK changes

PR #136, opened 2020-05-06
Add name filtering to the list beta group command

adamjcampbell:
Adds the ability to specify group names (or partial group names) for filtering down the beta group list

# 📝 Summary of Changes

- Adds `filterNames` (`--filter-names`) to `ListBetaGroupsCommand`
- Implements the filter by passing the argument through to the operation and endpoint

# ⚠️ Items of Note
It is interesting that the matching is partial and case insensitive
Something to keep in mind if re-using this operation as well

# 🧐🗒 Reviewer Notes

## 💁 Example

`USAGE: asc testflight betagroup list [--api-issuer <uuid>] [--api-key-id <keyid>] [--csv] [--json] [--table] [--yaml] [--filter-app-ids <app-id> ...] [--filter-bundle-ids <bundle-id> ...] [--filter-names <filter-names> ...]`

## 🔨 How To Test

- Run the list command with no filters
- Run the list command with the filter name argument with one of the previously displayed names to narrow the results

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/Operations/ListBetaGroupsOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/ListBetaGroupsOperation.swift
@@ -8,6 +8,7 @@ struct ListBetaGroupsOperation: APIOperation {
 
     struct Options {
         let appIds: [String]
+        let names: [String]
     }
 
     typealias BetaGroup = AppStoreConnect_Swift_SDK.BetaGroup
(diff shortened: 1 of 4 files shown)


Sources/AppStoreConnectCLI/Services/Operations/ListBetaGroupsOperation.swift
```diff
@@ -33,7 +34,10 @@ struct ListBetaGroupsOperation: APIOperation {
     }
 
     func execute(with requestor: EndpointRequestor) -> AnyPublisher<Output, Swift.Error> {
-        let filters = options.appIds.isEmpty ? [] : [ListBetaGroups.Filter.app(options.appIds)]
+        var filters: [ListBetaGroups.Filter] = []
+        filters += options.appIds.isEmpty ? [] : [.app(options.appIds)]
+        filters += options.names.isEmpty ? [] : [.name(options.names)]
+
```
DechengMa, line 40: not relevant to this PR, but after adding name as a filter I think the `ListBetaGroupOperation` should throw some Error if there are no groups in response. 🤔

adamjcampbell: 🤔 I'm not sure about that. In the list operation we're just filtering down the list. If the list happens to be empty after filtering that's okay I think.
The user will instinctively know to ask for the list with less filtering.

DechengMa: It will show a list with one empty line on it. I am not sure coz we seem did some error thrown in other listing commands when `data.count == 0` 🤔

adamjcampbell: Oh 🤔 we should be consistent on this. I think showing empty data is okay and doesn't need an error?

What do you think?

PR #137, opened 2020-05-06
List Builds Command

Nafisa2020:
# 📝 Summary of Changes

- lists builds of apps in App Store Connect

# ⚠️ Items of Note
Beta Group details are not included in the output as beta group is not returned as an included group from Builds API

## 🔨 How To Test
```swift run asc testflight builds list```

Screen Shot 2020-05-06 at 3 59 29 pm

```swift run asc testflight builds list --bundle-id iba.test3```
Screen Shot 2020-05-06 at 4 01 50 pm

```swift run asc testflight builds list --expired true```
Screen Shot 2020-05-06 at 4 02 40 pm

```swift run asc testflight builds list --build-number 5 10```
Screen Shot 2020-05-06 at 4 03 39 pm

### CODE REVIEW

File: Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/ListBuildsCommand.swift [hunk 1]
```diff
@@ -8,34 +8,56 @@ import Foundation
...
     @OptionGroup()
     var common: CommonOptions
 
-    @Argument(help: "A bundle identifier that uniquely identifies an application.")
-    var bundleId: String
+    @Option(
+      parsing: .upToNextOption,
+      help: "An opaque resource ID that uniquely identifies the build"
+    )
+    var bundleId: [String]
```
adamjcampbell, line 20: We have been prefixing all our filters with the `filter` keyword and specifying the value name as well

e.g.
```
    @Option(
        parsing: .upToNextOption,
        help: ArgumentHelp(
            "Filter by app bundle identifier. eg. com.example.App",
            discussion: "This option is mutually exclusive with --filter-app-ids.",
            valueName: "bundle-id"
        )
    ) var filterBundleIds: [String]
```

Also the help text provided here is actually the help text for an app id not a bundle id

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -177,6 +177,20 @@ class AppStoreConnectService {
         _ = try expireBuildOperation.execute(with: requestor).await()
     }
 
+  func listBuilds(bundleId: [String], expired: [String], preReleaseVersion: [String], buildNumber: [String], betaReviewState: [String], limit: Int?) throws -> [Build] {
+
+      var appId: [String] = []
```
adamjcampbell, line 182:
```suggestion
      var appIds: [String] = []
```

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,88 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct ListBuildsOperation: APIOperation {
+
+  struct Options {
+    let appId: [String]
```
adamjcampbell, line 10: All the filter arguments should be prefixed with filter

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,88 @@
...
+  init(options: Options) {
+    self.options = options
+  }
+
+  func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<[Build], Error> {
+
+    var filters = [ListBuilds.Filter]()
+
+    if !options.appId.isEmpty {
+      filters += [ListBuilds.Filter.app(options.appId)]
```
adamjcampbell, line 40: I have been reducing this to one line by using a ternary:
```swift
filters += options.filterAppIds.isEmpty ? [] : [ListBuilds.Filter.app(options.filterAppIds)]
```
Minimally less efficient but it does save some space

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,82 @@
...
+            include: [.app, .betaAppReviewSubmission, .buildBetaDetail, .preReleaseVersion],
+            limit: limit,
+            sort: [ListBuilds.Sort.uploadedDateAscending]
+        )
+
+
+        return requestor.request(endpoint)
+            .tryMap { (buildResponse) throws -> [Build] in
+                guard !buildResponse.data.isEmpty else {
+                    throw ListBuildsError.noBuildExist
```
adamjcampbell, line 71: Getting no results from a list endpoint is okay I think
I don't think an error is necessary for it they'll just get returned an empty list

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,82 @@
...
+                }
+
+                let buildDetailsInfo = buildResponse.data.map {
+                    Build($0, buildResponse.included)
```
adamjcampbell, line 75: We should construct an output here like we did for the Read command so we can return SDK models.
It can also be as simple as a `typealias` see: `ListBetaGroupsOperation`

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,61 @@
...
+            sort: [ListBuilds.Sort.uploadedDateAscending]
+        )
+
+        return requestor.request(endpoint)
+            .map { (buildResponse) -> Output in
+                return buildResponse.data.map {
+                    return ($0, buildResponse.included)
+                }
+        }
+        .eraseToAnyPublisher()
```
adamjcampbell, line 59: The indentation seems wonky here:
```suggestion
        return requestor.request(endpoint)
            .map { (buildResponse) -> Output in
                return buildResponse.data.map {
                    return ($0, buildResponse.included)
                }
            }
            .eraseToAnyPublisher()
```

There are a bunch of things you can omit here too if you so choose:
```suggestion
        return requestor.request(endpoint)
            .map { response -> Output in
                response.data.map { ($0, response.included) }
            }
            .eraseToAnyPublisher()
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -239,6 +239,39 @@ class AppStoreConnectService {
...
+                filterBuildNumbers: filterBuildNumbers,
+                filterProcessingStates: filterProcessingStates,
+                filterBetaReviewStates: filterBetaReviewStates,
+                limit: limit
+            )
+        )
+
+        let output = try listBuildsOperation.execute(with: requestor).await()
+        let builds = output.map { Build($0.build, $0.relationships) }
+        return builds
```
adamjcampbell, line 272: You can probably get away with a function reference here:
```suggestion
        return output.map(Build.init)
```

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,61 @@
...
+        filters += options.filterExpired.isEmpty ? [] : [ListBuilds.Filter.expired(options.filterExpired)]
+        filters += options.filterProcessingStates.isEmpty ? [] : [ListBuilds.Filter.processingState(options.filterProcessingStates)]
+        filters += options.filterBetaReviewStates.isEmpty ? [] :  [ListBuilds.Filter.betaAppReviewSubmissionBetaReviewState(options.filterBetaReviewStates)]
+
+        var limit: [ListBuilds.Limit]?
+
+        if let optionLimit = options.limit {
+            limit = [ListBuilds.Limit.individualTesters(optionLimit),
+                     ListBuilds.Limit.betaBuildLocalizations(optionLimit)]
+        }
```
adamjcampbell, line 44: This could be simplified to a map if you want:
```suggestion
        let limit = options.limit.map { limit -> [ListBuilds.Limit] in
            [.individualTesters(limit), .betaBuildLocalizations(limit)]
        }
```

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    private let options: Options
+
+    init(options: Options) {
+        self.options = options
+    }
+
+    func execute(with requestor: EndpointRequestor) -> AnyPublisher<Output, Error> {
+        var filters = [ListBuilds.Filter]()
+        filters += options.filterAppIds.isEmpty ? [] : [ListBuilds.Filter.app(options.filterAppIds)]
```
adamjcampbell, line 32: Relying on type inference you could choose to omit all the type prefixes here:
```suggestion
        filters += options.filterAppIds.isEmpty ? [] : [.app(options.filterAppIds)]
```

Issue #113, opened 2020-04-22
Read Build Information


orj:
# 🗣 Context

Get information about a specific build.

# 💬 Narrative

As a user
I want read the information of a specific build
So that I can inspect it

# 📝 Notes

Covers this [API](https://developer.apple.com/documentation/appstoreconnectapi/read_build_information).

# 🎨 Design

```
asc testflight build read bundleId version buildNumber 
```

Output:

Bundle Id | Name | Platform | Version | Build Number | Beta Review State | Processing State | External Build State | Internal Build State | Auto Notify | Min OS Version | Uploaded Date | Expiration Date | Expired | Uses Non Exempt Encryption | Beta Groups
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

# ✅ Acceptance Criteria

**GIVEN** ...
**WHEN** ...
**THEN** ...

PR #139, opened 2020-05-06
Expire a build

Nafisa2020:
# 📝 Summary of Changes

- expire a build

# ⚠️ Items of Note
- once a build has been expired it can't be undone
- As per the discussion modifying the usesNonExemptEncryption will be done later

## 🔨 How To Test

- Do a read before executing an expire command
   ```swift run asc testflight builds read iba.test2 ```

Screen Shot 2020-05-06 at 4 37 34 pm

- Execute the expire command as follows
    ```swift run asc testflight builds expire iba.test2 1.0 6 true```

- Check the change by executing another read command
    ```swift run asc testflight builds read iba.test2 ```

Screen Shot 2020-05-06 at 4 36 47 pm

### DIFF
--- a/Sources/AppStoreConnectCLI/Model/BuildDetailsInfo.swift
+++ b/Sources/AppStoreConnectCLI/Model/BuildDetailsInfo.swift
@@ -5,6 +5,7 @@ import Foundation
 import SwiftyTextTable
 
 struct BuildDetailsInfo: ResultRenderable {
+  let id: String
   let app: App?
   let platform: String?
   let version: String?
@@ -63,6 +64,7 @@ extension BuildDetailsInfo {
     let app = appDetails.map(App.init)
 
     self.init(
+      id: build.id,
       app: app,
       platform: prereleaseVersion?.attributes?.platform?.rawValue,
       version: prereleaseVersion?.attributes?.version,
--- a/Sources/AppStoreConnectCLI/Model/BuildDetailsInfo.swift
+++ b/Sources/AppStoreConnectCLI/Model/BuildDetailsInfo.swift
@@ -5,7 +5,6 @@ import Foundation
 import SwiftyTextTable
 
 struct BuildDetailsInfo: ResultRenderable {
-  let id: String
   let app: App?
   let platform: String?
   let version: String?
@@ -64,7 +63,6 @@ extension BuildDetailsInfo {
     let app = appDetails.map(App.init)
 
     self.init(
-      id: build.id,
       app: app,
       platform: prereleaseVersion?.attributes?.platform?.rawValue,
       version: prereleaseVersion?.attributes?.version,
--- a/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
+++ b/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
@@ -170,12 +170,14 @@ class AppStoreConnectService {
       let appId = try appsOperation.execute(with: requestor).await().map(\.id)
 
       let readBuildOperation = ReadBuildOperation(options: .init(appId: appId, buildNumber: [buildNumber], preReleaseVersion: [preReleaseVersion]))
-
       let buildId = try readBuildOperation.execute(with: requestor).await().map(\.id).first
+
       let modifyBuildOperation = ModifyBuildOperation(options: .init(buildId: buildId, expired: expired, usesNonExemptEncryption: usesNonExemptEncyption))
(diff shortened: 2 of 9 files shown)


Sources/AppStoreConnectCLI/Model/BuildDetailsInfo.swift
```diff
@@ -5,6 +5,7 @@ import Foundation
 import SwiftyTextTable
 
 struct BuildDetailsInfo: ResultRenderable {
+  let id: String
```
adamjcampbell, line 8: For our models we don't really want to expose the internal ids
`ReadBuildOperation` should return the SDK model so that you can access the `id`
Conversion to `BuildDetailsInfo` can be done as needed in the service

Sources/AppStoreConnectCLI/Services/Operations/ExpireBuildOperation.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+import Foundation
+
+struct ExpireBuildOperation: APIOperation {
+
+  struct Options {
+    let buildId: String?
```
adamjcampbell, line 10: Build id should be non optional ideally
I would expect`ReadBuildOperation` to return a single non optional model
It should take in one build number according to #113

Sources/AppStoreConnectCLI/Services/Operations/ExpireBuildOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+    let expired: Bool
```
adamjcampbell, line 11: We can delete this and just use `true` in the `execute` function

Sources/AppStoreConnectCLI/Services/Operations/ExpireBuildOperation.swift
```diff
@@ -0,0 +1,55 @@
...
+      usesNonExemptEncryption: nil)
+
+    return requestor.request(buildModifyEndpoint)
+      .tryMap { (buildResponse) throws in
+
+        guard (buildResponse.data.attributes != nil) else {
+          throw ExpireBuildError.noBuildExist
+        }
+    }
+    .eraseToAnyPublisher()
```
adamjcampbell, line 53: I'm not sure we need to throw errors here
If the modification is not successful the API will return a non `200` response code which the SDK will translate into an error

```suggestion
    return requestor.request(buildModifyEndpoint)
      .map { _ in }
      .eraseToAnyPublisher()
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -156,13 +156,25 @@ class AppStoreConnectService {
         return BetaGroup(app, betaGroup)
     }
 
-    func readBuild(bundleId: String, buildNumber: [String], preReleaseVersion: [String]) throws -> [BuildDetailsInfo] {
+    func readBuild(bundleId: String, buildNumber: String, preReleaseVersion: String) throws -> BuildDetailsInfo {
       let appsOperation = GetAppsOperation(options: .init(bundleIds: [bundleId]))
-      let appId = try appsOperation.execute(with: requestor).await().map(\.id)
+      let appId = try appsOperation.execute(with: requestor).await().map(\.id).first ?? ""
```
adamjcampbell, line 161:
```suggestion
      let appId = try appsOperation.execute(with: requestor).compactMap(\.first).await().id
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -156,13 +156,25 @@ class AppStoreConnectService {
...
       let readBuildOperation = ReadBuildOperation(options: .init(appId: appId, buildNumber: buildNumber, preReleaseVersion: preReleaseVersion))
 
-      return try readBuildOperation.execute(with: requestor).await()
+      let output = try readBuildOperation.execute(with: requestor).await()
+      return BuildDetailsInfo(output.build, output.relationships)
+    }
+
+    func expireBuild(bundleId: String, buildNumber: String, preReleaseVersion: String) throws -> Void {
+      let appsOperation = GetAppsOperation(options: .init(bundleIds: [bundleId]))
+      let appId = try appsOperation.execute(with: requestor).await().map(\.id).first ?? ""
```
adamjcampbell, line 171:
```suggestion
      let appId = try appsOperation.execute(with: requestor).compactMap(\.first).await().id
```

Issue #92, opened 2020-04-22
Bug: No paging support


orj:
# 💬 Description

Currently none of our commands support paging.

# 🦶 Reproduction Steps

Steps to reproduce the behavior, provide an example of the command line issued:
1. Run a command on a dataset that exceeds the default page limit.

# 🤔 Expected Results

That all pages of data are fetched.

# 😲 Actual Results

On the first page of data is fetched.

orj:
We can probably embed paging into the domain layer in #41

PR #141, opened 2020-05-11
Add pagination for List Build Command

DechengMa:
Add pagination support for List Build Command, the ultimate goal is closing #92 

#### Background: 
Apple store connect API pagination is based on this field in the response. which will point to the next page
```
"links": {
    "self": "https://api.appstoreconnect.apple.com/v1/builds?cursor=MG.ABCDE",
    "first": "https://api.appstoreconnect.apple.com/v1/builds"
},
```
and SDK has links field `PagedDocumentLinks`, input this to APIEndpoint can fetch more.

# 📝 Summary of Changes

- Create `PaginatedResponse` 
- Create a function `concatFetcher` extends `EndpointRequestor`, for **recursive** fetching rest items.
- Tweak list build Operation for demo  

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage: `swift run asc testflight builds list`
Sample Response : 
`All builds will be shown ` 

## 🔨 How To Test

`swift run asc testflight builds list`

## TODO
the default limit for fetching items is **50**, we can change to **200** for reducing requests

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
+++ b/Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
@@ -24,3 +24,25 @@ struct DefaultEndpointRequestor: EndpointRequestor {
         }
     }
 }
+
+protocol PaginatedResponse: Decodable {
+    var links: PagedDocumentLinks { get }
+}
+
+extension EndpointRequestor {
+    func concatFetcher<T: PaginatedResponse>(_ endpointMaker: @escaping (PagedDocumentLinks?) -> APIEndpoint<T>, next: PagedDocumentLinks?) -> AnyPublisher<[T], Error> {
+        self.request(endpointMaker(next)).flatMap { (response) -> AnyPublisher<[T], Error> in
+            if response.links.next != nil {
+                return self.concatFetcher(endpointMaker, next: response.links)
+                    .flatMap {
+                        Empty<[T], Error>()
+                            .append([response] + $0)
+                    }
+                    .eraseToAnyPublisher()
+            }
+
+            return Empty<[T], Error>().append([response]).eraseToAnyPublisher()
+        }
+        .eraseToAnyPublisher()
+    }
+}
--- a/Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
+++ b/Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
@@ -30,22 +30,18 @@ protocol PaginatedResponse: Decodable {
 }
 
 extension EndpointRequestor {
-    func concatFetcher<T: PaginatedResponse>(
-        with endpointMaker: @escaping (PagedDocumentLinks?) -> APIEndpoint<T>,
-        next: PagedDocumentLinks?
+    func requestAllPages<T: PaginatedResponse>(
+        with endpointProvider: @escaping (PagedDocumentLinks?) -> APIEndpoint<T>,
+        next: PagedDocumentLinks? = nil
     ) -> AnyPublisher<[T], Error> {
-        self.request(endpointMaker(next)).flatMap { (response) -> AnyPublisher<[T], Error> in
-            if response.links.next != nil {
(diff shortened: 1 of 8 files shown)


Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
```diff
@@ -24,3 +24,28 @@ struct DefaultEndpointRequestor: EndpointRequestor {
...
+                            .append([response] + $0)
+                    }
+                    .eraseToAnyPublisher()
+            }
+
+            return Empty<[T], Error>().append([response]).eraseToAnyPublisher()
+        }
+        .eraseToAnyPublisher()
+    }
+}
```
adamjcampbell, line 51: 🤔 I think we can make a couple small improvements here
What do you think of the function name: `requestAllPages`

Additionally we can default `next` to nil if not provided (next: PagedDocumentLinks? = nil)

Rather than the `Maker` suffix we usually see the `Provider` suffix. Or we might even bet able to get away with just naming it `endpoint` as a function

Also if we flip the parameters to provide the closure second we can have the option to use trailing closure syntax

Sources/AppStoreConnectCLI/Services/EndpointRequestor.swift
```diff
@@ -24,3 +24,24 @@ struct DefaultEndpointRequestor: EndpointRequestor {
...
+        with endpointProvider: @escaping (PagedDocumentLinks?) -> APIEndpoint<T>,
+        next: PagedDocumentLinks? = nil
+    ) -> AnyPublisher<[T], Error> {
+        request(endpointProvider(next))
+            .flatMap { (response) -> AnyPublisher<[T], Error> in
+                response.links.next != nil
+                    ? self.requestAllPages(with: endpointProvider, next: response.links)
+                        .flatMap { Empty<[T], Error>().append([response] + $0) }
+                        .eraseToAnyPublisher()
+                    : Empty<[T], Error>().append([response]).eraseToAnyPublisher()
```
adamjcampbell, line 43: An alternative here is:
```swift
Just([response]).setFailureType(to: Error.self).eraseToAnyPublisher()
```

adamjcampbell, line 43: Even though it's multi return a guard is probably clearer here rather than a large ternery

```suggestion
                guard response.links.next != nil else {
                    return Empty<[T], Error>().append([response]).eraseToAnyPublisher()
                }
                
                return self.requestAllPages(with: endpointProvider, next: response.links)
                    .flatMap { Empty<[T], Error>().append([response] + $0) }
                    .eraseToAnyPublisher()
```

What do you think?

Sources/AppStoreConnectCLI/Services/Operations/ListBuildsOperation.swift
```diff
@@ -36,21 +29,38 @@ struct ListBuildsOperation: APIOperation {
...
+                next: $0
+            )
+        }
+        .map { (responses: [BuildsResponse]) -> Output in
+            responses.flatMap { (response: BuildsResponse) -> Output in
+                (response.data.map { ($0, response.included) })
             }
-            .eraseToAnyPublisher()
+        }
+        .eraseToAnyPublisher()
```
adamjcampbell, line 62: Might be more readable if we pull out the endpoint params, then we can put the endpoint provider function on one line

```swift
let filter = self.filters
let include: [...] = [.app, .betaAppReviewSubmission, .buildBetaDetail, .preReleaseVersion]
let limit = self.limit
let sort: [...] = [.uploadedDateAscending]

return requestor
    .requestAllPages { .builds(filter: filters, include: includes, limit: limit, sort: sort, next: $0) }
    .map { (responses: [BuildsResponse]) -> Output in
        responses.flatMap { (response: BuildsResponse) -> Output in
            (response.data.map { ($0, response.included) })
        }
    }
    .eraseToAnyPublisher()
```

PR #143, opened 2020-05-13
Implement Add build to groups command

### LINKED ISSUE #114: Add Access for Beta Groups to a Build
# 🗣 Context

Add or create a beta group to a build to enable testing.

# 💬 Narrative

As a user
I want make a build available to beta groups
So that so that beta testers can test the build

# 📝 Notes

Covers this [API](https://developer.apple.com/documentation/appstoreconnectapi/add_access_for_beta_groups_to_a_build).

# 🎨 Design

```
asc testflight build addbetagroup bundleId version buildNumber groups...
```

# ✅ Acceptance Criteria

**GIVEN** ...
**WHEN** ...
**THEN** ...

DechengMa:
Closes #114 

# 📝 Summary of Changes

- Create `BuildArguments` for reuse arguments 
- Implement `AddGroupsToBuildCommand` and `AddGroupsToBuildOperation`
- Add paingation to ListBetaGroupsOperation 
- Use local filter when finding groups.

# ⚠️ Items of Note

Can discuss if extract argument to OptionGroup is needed 

# 🧐🗒 Reviewer Notes

## 💁 Example

USAGE: 
`asc testflight builds addbetagroup <bundle-id> <pre-release-version> <build-number> [<group-names> ...]`

## 🔨 How To Test

`asc testflight builds addbetagroup com.example.foo 1.0.0 2 TestGroup`

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
+++ b/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
@@ -405,31 +405,32 @@ class AppStoreConnectService {
         buildNumber: String,
         groupNames: [String]
     ) throws {
-        let appId = try ReadAppOperation(options: .init(identifier: .bundleId(bundleId)))
-            .execute(with: requestor)
-            .await()
-            .id
-
-        let buildId = try ReadBuildOperation(
-                options: .init(
-                    appId: appId,
-                    buildNumber: buildNumber,
-                    preReleaseVersion: version
-                )
-            )
-            .execute(with: requestor)
-            .await()
-            .build
-            .id
-
-        let groupIds = try groupNames.flatMap {
-            try ListBetaGroupsOperation(options: .init(appIds: [appId], names: [$0]))
-                .execute(with: requestor)
-                .await()
-                .map(\.betaGroup.id)
-        }
+        let (buildId, groupIds) = try getBuildIdAndGroupIdsFrom(
+            bundleId: bundleId,
+            version: version,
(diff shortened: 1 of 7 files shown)


Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -376,3 +377,40 @@ class AppStoreConnectService {
...
+            .await()
+            .build
+            .id
+
+        let groupIds = try groupNames.flatMap {
+            try ListBetaGroupsOperation(options: .init(appIds: [appId], names: [$0]))
+                .execute(with: requestor)
+                .await()
+                .map(\.betaGroup.id)
+        }
```
adamjcampbell, line 411: Why not something like:
```suggestion
        let groupIds = try ListBetaGroupsOperation(options: .init(appIds: [appId], names: groupNames))
                .execute(with: requestor)
                .await()
                .map(\.betaGroup.id)
```

I feel we could make a single network request here rather than many

DechengMa: After some tries found API will aggregate all the filters so input two filters at a time: 
`name1 && name2` can find a group called ->  `name1name2` 
Seems multiple requests is required when searching with multiple group names 🤔

adamjcampbell: I feel like with n group names it might be less requests to load all groups for a single app and then filter in the tool...
Let's try that

PR #147, opened 2020-05-18
Implement Sync Beta Group Commands

DechengMa:
Implement basic sync beta group command.

- [x] Delete local group files -> request to remove groups in server
- [x] Create local group files -> request create groups in server 
- [x] After create -> pull back and sync local
- [x] Edit local group files -> update server groups 
- [x] Wrap up and restructure codes
- [x] Apply new file system on it
- [ ] Refactor file structure and remove duplicate data

# 📝 Summary of Changes

- Add `let id: String?` field to Beta Group model. I know we don't want to do this but it's a must coz two groups can have the same name.
- Add `var testers: [String]` to Beta Group for storing tester emails
- Make `BetaGroup` extend to `Equatable, Hashable` for comparing in `Set`, to `SyncResultRenderable` for rendering dry run and sync result, to `FileNameGettable` for generating a beatified filename (a String extension was made for snakeCased filename). 
- Create `SyncResultRenderer` for printing sync and dry run result 
- Create `enum SyncStrategy` for both rendering result and handle the update.
- Create `protocol ResourceFolderManager`and `BetaGroupFolderManager`, for managing folders and group files.
- Tweak `ListBetaTestersOperation` and create `UpdateBetaGroupOperation`. Create service function `pullBetaGroups`, `updateBetaGroup`, `deleteBetaGroup`. 
- Implement `SyncBetaGroupsCommand` which contains `PullBetaGroupsCommand` and `PushBetaGroupsCommand`.

# ⚠️ Items of Note
- When edit file to change group, only particular field are editable: for a beta group, only name, public link, link limit, link enabled can be changed (according to API), and if others are changed, when push it will have no effect. So this means the user must have some instructions.

# 🧐🗒 Reviewer Notes

#### Usage: 
`asc testflight betagroup sync pull [--output-path <output-path>]`
`asc testflight betagroup sync push [--input-path <input-path>] [--dry-run]`

`--output-path` and `--input-path` is a folder path. 

CLI will go through all the files in that folder and find file with a suffix of `yml` and read it as a beta group file.

## 🔨 How To Test

`asc testflight betagroup sync pull`

`asc testflight betagroup sync push --dry-run`

`asc testflight betagroup sync push`

#### Create a new group via sync: 
1. Create a file, name it whatever you like in folder `./config/betagroups`(or the input folder path), suffix **must** be `yml`
2. Put some basic group info in it, eg
```
app:
  bundleId: com.example.foo // MUST HAVE
groupName: testgroup // MUST HAVE, MUST UNIQUE (please don't create group with the same name in same app)
publicLinkEnabled: false
publicLinkLimit: 5
publicLinkLimitEnabled: false
```
3. Run command `asc testflight betagroup sync push`

#### Delete a group 
1. Delete the group file in folder 
2. Run command `asc testflight betagroup sync push`

#### Update a group 
1. Directly edit the group file in the config folder 
fields that support editing: (edit other fields will not work)
```
groupName
publicLinkEnabled
publicLinkLimit
publicLinkLimitEnabled
```
2. Run command `asc testflight betagroup sync push`

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
@@ -1,0 +1,30 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import FileSystem
+
+struct PullBetaGroupsCommand: CommonParsableCommand {
+
+    static var configuration = CommandConfiguration(
+        commandName: "pull",
+        abstract: "Pull down server beta groups, refresh local beta group config files"
+    )
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Option(
+        default: "./config/betagroups",
+        help: "Path to the Folder containing the information about beta groups. (default: './config/betagroups')"
+    ) var outputPath: String
+
+    func run() throws {
+        let service = try makeService()
+
+        let betaGroupWithTesters = try service.pullBetaGroups()
+
+        try BetaGroupProcessor(path: .folder(path: outputPath))
+            .write(groupsWithTesters: betaGroupWithTesters)
+    }
+
+}
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
@@ -7,7 +7,7 @@ struct PullBetaGroupsCommand: CommonParsableCommand {
 
     static var configuration = CommandConfiguration(
         commandName: "pull",
-        abstract: "Pull down server beta groups, refresh local beta group config files"
+        abstract: "Pull down existing beta groups, refreshing local beta group config files"
     )
 
     @OptionGroup()
@@ -15,7 +15,7 @@ struct PullBetaGroupsCommand: CommonParsableCommand {
 
     @Option(
         default: "./config/betagroups",
-        help: "Path to the Folder containing the information about beta groups. (default: './config/betagroups')"
+        help: "Path to the Folder containing the information about beta groups."
     ) var outputPath: String
 
     func run() throws {
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
@@ -1,6 +1,8 @@
 // Copyright 2020 Itty Bitty Apps Pty Ltd
 
 import ArgumentParser
+import FileSystem
+import Foundation
 import struct Model.BetaGroup
 
 struct PushBetaGroupsCommand: CommonParsableCommand {
@@ -22,7 +24,66 @@ struct PushBetaGroupsCommand: CommonParsableCommand {
     var dryRun: Bool
 
     func run() throws {
+        let service = try makeService()
 
+        let resourceProcessor = BetaGroupProcessor(path: .folder(path: inputPath))
+
+        let serverGroups = Set(try service.pullBetaGroups().map{ $0.betaGroup })
+        let localGroups = Set(try resourceProcessor.read())
+
+        let strategies = compareGroups(
+            localGroups: localGroups,
+            serverGroups: serverGroups
+        )
+
+        let renderer = Renderers.SyncResultRenderer<BetaGroup>()
+
+        if dryRun {
+            renderer.render(strategies, isDryRun: true)
+        } else {
+            try strategies.forEach { (strategy: SyncStrategy) in
+                switch strategy {
+                case .create(let group):
+                    _ = try service.createBetaGroup(
+                        appBundleId: group.app.bundleId!,
+                        groupName: group.groupName,
+                        publicLinkEnabled: group.publicLinkEnabled ?? false,
+                        publicLinkLimit: group.publicLinkLimit
+                    )
+                case .delete(let group):
+                    try service.deleteBetaGroup(with: group.id!)
+                case .update(let group):
+                    try service.updateBetaGroup(betaGroup: group)
+                }
+
+                renderer.render(strategy, isDryRun: false)
+            }
+
+            let betaGroupWithTesters = try service.pullBetaGroups()
+
+            try resourceProcessor.write(groupsWithTesters: betaGroupWithTesters)
+        }
+    }
+
+    func compareGroups(localGroups: Set<BetaGroup>, serverGroups: Set<BetaGroup>) -> [SyncStrategy<BetaGroup>] {
+        var strategies: [SyncStrategy<BetaGroup>] = []
+
+        let groupToCreate = localGroups.subtracting(serverGroups)
+        let groupToDelete = serverGroups.subtracting(localGroups)
+
+        groupToDelete.forEach { group in
+            if !localGroups.contains(where: { group.id == $0.id }) {
+                strategies.append(.delete(group))
+            }
+        }
+
+        groupToCreate.forEach { group in
+            serverGroups.contains(where: { group.id == $0.id })
+                ? strategies.append(.update(group))
+                : strategies.append(.create(group))
+        }
+
+        return strategies
     }
 
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
@@ -1,0 +1,28 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import struct Model.BetaGroup
+
+struct PushBetaGroupsCommand: CommonParsableCommand {
+
+    static var configuration = CommandConfiguration(
+        commandName: "push",
+        abstract: "Push local beta group config files to server, update server beta groups"
+    )
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Option(
+        default: "./config/betagroups",
+        help: "Path to the Folder containing the information about beta groups. (default: './config/betagroups')"
+    ) var inputPath: String
+
+    @Flag(help: "Perform a dry run.")
+    var dryRun: Bool
+
+    func run() throws {
+
+    }
+
+}
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
@@ -28,13 +28,14 @@ struct PushBetaGroupsCommand: CommonParsableCommand {
 
         let resourceProcessor = BetaGroupProcessor(path: .folder(path: inputPath))
 
-        let serverGroups = Set(try service.pullBetaGroups().map { $0.betaGroup })
-        let localGroups = Set(try resourceProcessor.read())
+        let serverGroups = try service.pullBetaGroups().map { $0.betaGroup }
+        let localGroups = try resourceProcessor.read()
 
-        let strategies = compareGroups(
-            localGroups: localGroups,
-            serverGroups: serverGroups
-        )
+        let strategies = SyncResourceComparator(
+                localResources: localGroups,
+                serverResources: serverGroups
+            )
+            .compare()
 
         let renderer = Renderers.SyncResultRenderer<BetaGroup>()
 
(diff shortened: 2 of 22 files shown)


Sources/AppStoreConnectCLI/Model/BetaGroup.swift
```diff
@@ -62,6 +65,37 @@ extension BetaGroup {
...
+
+extension BetaGroup: SyncResultRenderable {
+    var syncResultText: String {
+        self.groupName ?? "group-name"
+    }
+}
+
+extension BetaGroup: FileNameGettable {
+    var fileName: String {
+        "\(app.name ?? "")_\(groupName ?? "group-name")"
```
DechengMa, line 95: using app name at first for differentiate, because every app will have a default test group called `App store connect user` and one will override the other.

Sources/AppStoreConnectCLI/Services/Operations/ListBetaTestersOperation.swift
```diff
@@ -103,12 +92,8 @@ struct ListBetaTestersOperation: APIOperation {
         }
 
         return requestor.request(endpoint)
-            .tryMap { (response: BetaTestersResponse) -> Output in
-                guard !response.data.isEmpty else {
-                    throw Error.notFound
-                }
```
DechengMa, line 109: Found It's ok for list command to have zero result.

Sources/AppStoreConnectCLI/Files/BetaGroupFileManager.swift
```diff
@@ -0,0 +1,58 @@
...
+                .createFile(named: "\(group.formattedFileName)_beta-testers.csv")
+
+            try testersFile.write(testers.renderAsCSV())
+
+            var group = group
+
+            group.testers = testersFile.path
+
+            return group
+        }
```
adamjcampbell, line 54: I can see what you were going for here, however you usually don't want to perform side effects in a `map` function. To explain more explicitly you would usually expect your `map` to be a pure transformation and not do things like write out files.

This could be cleaned up by extracting a private func `save(_ betaGroup:)` out of `save(_ betaGroups:)` and use a for each loop here to write out the testers and then the betagroup.

Now that I write this though it occurs to me that we probably want a singular file to describe all the groups? But maybe not... One file per group might be okay. @orj any thoughts?

Sources/AppStoreConnectCLI/Files/ResourceFolderManager.swift
```diff
@@ -0,0 +1,26 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+protocol ResourceFolderManager {
```
adamjcampbell, line 5: Like many developers I have grown an aversion to the work 'Manager' how about `ResourceProcessor`

Also I feel the function name `read` pairs better with the word `write` as opposed to `save` what do you think?

Sources/AppStoreConnectCLI/Files/ResourceFolderManager.swift
```diff
@@ -0,0 +1,26 @@
...
+    associatedtype T: Codable
+
+    var path: String { get }
+
+    func read() throws -> [T]
+
+    func save(_: [T]) throws
+}
+
+protocol FileNameGettable {
```
adamjcampbell, line 15: `Provider` is a little more idiomatic compared to `Gettable`
John Sundell wrote some good thoughts regarding different kinds of protocols here: https://www.swiftbysundell.com/articles/different-categories-of-swift-protocols/

Sources/AppStoreConnectCLI/Helpers/String+Helpers.swift
```diff
@@ -10,4 +10,18 @@ extension String {
         return String(String.UnicodeScalarView(filtered))
     }
 
+    func snakeCased() -> String? {
```
adamjcampbell, line 13: Any reason we need to bother with snake casing our names? This type of regex matching is prone to breaking when edge cases arise. We can't know how people will name their apps, betaGroups, etc

Sources/AppStoreConnectCLI/Model/App.swift
```diff
@@ -5,8 +5,8 @@ import Combine
 import Foundation
 import SwiftyTextTable
 
-struct App: ResultRenderable, Equatable {
-    let id: String
+struct App: ResultRenderable, Equatable, Hashable {
+    let id: String?
```
adamjcampbell, line 9: I believe `id` should never be nil if we have an App?

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
```diff
@@ -5,7 +5,8 @@ import Combine
 import Foundation
 import SwiftyTextTable
 
-struct BetaGroup: TableInfoProvider, ResultRenderable, Equatable {
+struct BetaGroup: TableInfoProvider, ResultRenderable {
+    let id: String?
```
adamjcampbell, line 9: Same goes for BetaGroup

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
```diff
@@ -14,6 +15,7 @@ struct BetaGroup: TableInfoProvider, ResultRenderable, Equatable {
     let publicLinkLimit: Int?
     let publicLinkLimitEnabled: Bool?
     let creationDate: String?
+    var testers: String? // CSV file path
```
adamjcampbell, line 18: This is where things can get interesting.. Having variables that will populated when pulled from disk but not populated when retrieved from the network

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
```diff
@@ -62,6 +65,37 @@ extension BetaGroup {
...
+            lhs.publicLinkLimitEnabled == rhs.publicLinkLimitEnabled
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(id)
+        hasher.combine(groupName)
+        hasher.combine(publicLinkEnabled)
+        hasher.combine(publicLinkLimit)
+        hasher.combine(publicLinkLimitEnabled)
+    }
```
adamjcampbell, line 84: We could potentially only use the id for matching.. do we ever foresee having duplicates?

Did you have to override the default implementations because of the testers file path? There may also be a different solution here 🤔

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
```diff
@@ -62,6 +65,37 @@ extension BetaGroup {
...
+}
+
+extension BetaGroup: SyncResultRenderable {
+    var syncResultText: String {
+        self.groupName ?? "group-name"
```
adamjcampbell, line 89: I feel like if the group has no name we have a serious problem. Maybe we need to tighten our model and errors for not having a name

Sources/AppStoreConnectCLI/Services/Operations/UpdateBetaGroupOperation.swift
```diff
@@ -0,0 +1,33 @@
...
+    init(options: Options) {
+        self.options = options
+    }
+
+    func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<BetaGroupResponse, Error> {
+        let betaGroup = options.betaGroup
+
+        let endpoint = APIEndpoint.modify(
+            betaGroupWithId: betaGroup.id!,
+            name: betaGroup.groupName!,
```
adamjcampbell, line 24: Again, if things are required we should look into them being non optional

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PushBetaGroupsCommand.swift
```diff
@@ -0,0 +1,84 @@
...
+                renderer.render(strategy, isDryRun: false)
+            }
+
+            let betaGroupWithTesters = try service.pullBetaGroups()
+
+            try folderManager.save(groupsWithTesters: betaGroupWithTesters)
+        }
+    }
+
+    func compareGroups(localGroups: Set<BetaGroup>, serverGroups: Set<BetaGroup>) -> [SyncStrategy<BetaGroup>] {
```
DechengMa, line 64: This can also be made generic, for other sync commands

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
DechengMa: Yeh, for now, this is only for comparing the BetaGroup object difference, not for testers. when doing something like `Set(localGroups).subtracting(serverGroups)` I can get the difference and update the server by local

Sources/AppStoreConnectCLI/Model/BetaGroup.swift
DechengMa: An optional id is for creating a beta group. When creating a new group we don't have the id yet. we can basically create a file with just minimal information (eg. group name, link enabled, etc) and push back

orj:
@DechengMa sorry I've not reviewed this. I'd like to try building a workflow based on this work to test it out.

DechengMa:
> @DechengMa sorry I've not reviewed this. I'd like to try building a workflow based on this work to test it out.

Sure no problem. Please PM directly if there's anything that I can do to help. And I'll be around for fixing any bugs / unexpected behaviours in the first place.

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
```diff
@@ -0,0 +1,30 @@
...
+        commandName: "pull",
+        abstract: "Pull down server beta groups, refresh local beta group config files"
+    )
+
+    @OptionGroup()
+    var common: CommonOptions
+
+    @Option(
+        default: "./config/betagroups",
+        help: "Path to the Folder containing the information about beta groups. (default: './config/betagroups')"
```
orj, line 18: The default will be printed automatically.
```suggestion
        help: "Path to the Folder containing the information about beta groups."
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/PullBetaGroupsCommand.swift
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import FileSystem
+
+struct PullBetaGroupsCommand: CommonParsableCommand {
+
+    static var configuration = CommandConfiguration(
+        commandName: "pull",
+        abstract: "Pull down server beta groups, refresh local beta group config files"
```
orj, line 10:
```suggestion
        abstract: "Pull down existing beta groups, refreshing local beta group config files."
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/SyncBetaGroupsCommand.swift
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+
+struct SyncBetaGroupsCommand: ParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "sync",
+        abstract: """
+        Sync information about beta groups with provided configuration file.
+        """,
```
orj, line 10: This doesn't need to be a multi-line literal.
```suggestion
        abstract: "Sync information about beta groups with provided configuration file.",
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaGroups/Sync/SyncBetaGroupsCommand.swift
```diff
@@ -0,0 +1,17 @@
...
+        subcommands: [
+            PullBetaGroupsCommand.self,
+            PushBetaGroupsCommand.self,
+        ],
+        defaultSubcommand: PullBetaGroupsCommand.self
```
orj, line 15: I don't think we should have a default command.
```suggestion
        ]
```

Sources/FileSystem/Beta Group/BetaGroupProcessor.swift
```diff
@@ -0,0 +1,66 @@
...
+            .write(group: group, testers: betaTesters)
+
+        return group
+    }
+
+}
+
+extension BetaGroup: FileProvider {
+    var fileName: String {
+        "\(app.bundleId ?? "")_\(groupName).yml"
```
orj, line 60: I think we should be normalising the filename a bit better here. `groupName` could be an unsafe filename.  We should turn all non alpha numeric characters into underscores.

```
extension String {
  func filenameSafe() -> String {
    let unsafeFilenameCharacters = CharacterSet(charactersIn: " *?:/\\.")
    return str.components(separatedBy: unsafeFilenameCharacters).joined(separator: "_")
  }
}
```

orj: For an account that has a lot of groups it would be better to nest them in subdirectories.  The `app.bundleId` could be a subdirectory rather than a prefix.

Also, surely the `app.bundleId` will be non-nil?  When will it ever be nil?

orj: Actually thinking about he normalised/safe filename some more I think when group names are translated to file/path names they should be really cleaned up:

Eg: A group name like `d_App Team (Learn & Adopt)` should be translated to `d-app-team-learn-adopt`. The problem with that however is that removing characters might mean there could be clashes. We might have to come up with some resolution strategy in that case.

orj: What I'm trying to suggest here is that the filenames should be names that a developer might choose if they were authoring them themselves.

orj:
Something else I'd like us to clean up here is the duplication of information. When we pull down the details of the beta group the `app` element is repeated in each betagroup file.  This isn't ideal. It would be better if it was not repeated.

DechengMa:
a new redesigned sync command will be implemented in #203

PR #149, opened 2020-05-21
Extract App and BetaGroup model into Model module

adamjcampbell:
This PR introduces a `Model` module to house all our models. We can then start sharing this module between different modules we create to start talking a common language.

# 📝 Summary of Changes

- Make `Model` target / module
- Move `App` to `Model`
- Move `BetaGroup` to `Model`

# ⚠️ Items of Note

We have a new module! Hopefully over time we can hide the SDK and SDK types behind a networking module of some kind so only that module needs to deal with two sets of models


Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -4,6 +4,9 @@ import AppStoreConnect_Swift_SDK
 import Combine
 import Foundation
 
+import struct Model.App
+import struct Model.BetaGroup
+
```
DechengMa, line 9: This list will be pretty long, we need network layer 🚁

Sources/AppStoreConnectCLI/Model/Build.swift
```diff
@@ -2,124 +2,125 @@
...
+                return betaAppReviewSubmission
+            }
+            return nil
+        }
 
-    let appDetails = includedApps?.filter { relationships?.app?.data?.id == $0.id }.first
-    let prereleaseVersion = includedPrereleaseVersions?.filter { relationships?.preReleaseVersion?.data?.id == $0.id }.first
-    let buildBetaDetail = includedBuildBetaDetails?.filter { relationships?.buildBetaDetail?.data?.id == $0.id }.first
-    let betaAppReviewSubmission = includedBetaAppReviewSubmissions?.filter { relationships?.betaAppReviewSubmission?.data?.id == $0.id }.first
+        let appDetails = includedApps?.filter { relationships?.app?.data?.id == $0.id }.first
```
Nafisa2020, line 59: Shouldn't we use first(where: ) here as we discussed yesterday for listprereleaseversion command?

Package.swift
```diff
@@ -43,20 +43,20 @@ let package = Package(
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
         // Targets can depend on other targets in this package, and on products in packages which this package depends on.
-        .target(
-            name: "appstoreconnect-cli",
-            dependencies: ["AppStoreConnectCLI"]),
+        .target(name: "appstoreconnect-cli", dependencies: ["AppStoreConnectCLI"]),
+        .target(name: "Model"),
```
DechengMa, line 47: Do we want to have a more implicit name here? like AppleStoreConnectCLI-Model?

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
adamjcampbell: Yeah when we move service / networking moves out we can just use `import Model`

Package.swift
adamjcampbell: I don't think so, these modules are only visible to our package as far as I understand

adamjcampbell: Also this package isn't built to be consumed, only to be compiled into the `asc` binary

Sources/AppStoreConnectCLI/Model/Build.swift
adamjcampbell: Potentially, however this was just a straight format of the file. I might raise a separate PR for it since it's not the focus of this PR

PR #153, opened 2020-05-22
Improve list builds command

orj:
This PR attempts to "improve" the UI on the the `testflight builds list` command.

# 📝 Summary of Changes

- It made sense to have the internal and external state columns reversed to me.
- And filtering by expired 'true' or 'false' felt weird.  So I switched it to a `include-expired` flag.

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

```
asc testflight builds list -h
asc testflight builds list
asc testflight builds list --include-expired
asc testflight builds list --no-include-expired
```


Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/ListBuildsCommand.swift
```diff
@@ -76,7 +77,7 @@ struct ListBuildsCommand: CommonParsableCommand {
 
         let builds = try service.listBuilds(
             filterBundleIds: filterBundleIds,
-            filterExpired: filterExpired,
+            filterExpired: includeExpired ? [] : ["false"],
```
adamjcampbell, line 80: This could be nicer 🤔 
But this is probably fine for now

PR #159, opened 2020-05-25
Extract the rest of the Models to Model module

adamjcampbell:
Provides more separation for our domain models for moving to a more modular architecture

# 📝 Summary of Changes

- Moves the rest of the Models in `AppStoreConnectCLI` to `Model`

# ⚠️ Items of Note

We lose the auto generated struct initialiser because it isn't public so we have to manually write it. Also I have not redefined the SDK enums so many things that were previously stored as enum are now `String` by passing their `rawValue`

### DIFF
--- a/Sources/AppStoreConnectCLI/Model/Build.swift
+++ b/Sources/AppStoreConnectCLI/Model/Build.swift
@@ -2,27 +2,10 @@
 
 import AppStoreConnect_Swift_SDK
 import Foundation
-import struct Model.App
+import Model
 import SwiftyTextTable
 
-struct Build: ResultRenderable {
-    let app: App?
-    let platform: String?
-    let version: String?
-    let externalBuildState: String?
-    let internalBuildState: String?
-    let autoNotifyEnabled: String?
-    let buildNumber: String?
-    let processingState: String?
-    let minOsVersion: String?
-    let uploadedDate: String?
-    let expirationDate: String?
-    let expired: String?
-    let usesNonExemptEncryption: String?
-    let betaReviewState: String?
-}
-
-extension Build {
+extension Model.Build {
 
     init(_ build: AppStoreConnect_Swift_SDK.Build ,_ includes: [AppStoreConnect_Swift_SDK.BuildRelationship]?) {
 
@@ -61,7 +44,7 @@ extension Build {
         let buildBetaDetail = includedBuildBetaDetails?.filter { relationships?.buildBetaDetail?.data?.id == $0.id }.first
         let betaAppReviewSubmission = includedBetaAppReviewSubmissions?.filter { relationships?.betaAppReviewSubmission?.data?.id == $0.id }.first
 
-        let app = appDetails.map(App.init)
+        let app = appDetails.map(Model.App.init)
 
         self.init(
             app: app,
@@ -82,7 +65,7 @@ extension Build {
     }
 }
 
-extension Build: TableInfoProvider {
+extension Model.Build: ResultRenderable, TableInfoProvider {
     static func tableColumns() -> [TextTableColumn] {
         return [
             TextTableColumn(header: "Bundle Id"),
@@ -104,7 +87,7 @@ extension Build: TableInfoProvider {
     }
 
     var tableRow: [CustomStringConvertible]  {
-        return [
+        let row = [
             app?.bundleId,
             app?.name,
             platform,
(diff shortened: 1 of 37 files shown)


adamjcampbell:
@Nafisa2020 The end goal is to move the SDK to a Network module so a lot of the imports will disappear as a result. There will be no conflicting imports at the Command level, which will be nice

You're right, it's a little messy but it should tidy up as we progress

Sources/AppStoreConnectCLI/Model/Build.swift
```diff
@@ -104,7 +87,7 @@ extension Build: TableInfoProvider {
     }
 
     var tableRow: [CustomStringConvertible]  {
-        return [
+        let row = [
```
Nafisa2020, line 90: Do we need this temp var row?

adamjcampbell: I was just attempting to make the code clearer / read nicer. In all honesty it's probably optimized out by the compiler. It's not strictly necessary, no.

Nafisa2020: It's just other places we are returning without taking it in a temp var row..didn't look consistent to me

Sources/AppStoreConnectCLI/Model/Profile.swift
```diff
@@ -2,33 +2,23 @@
...
             name: attributes.name,
-            platform: attributes.platform,
+            platform: attributes.platform?.rawValue,
             profileContent: attributes.profileContent,
             uuid: attributes.uuid,
             createdDate: attributes.createdDate,
-            profileState: attributes.profileState,
-            profileType: attributes.profileType,
+            profileState: attributes.profileState?.rawValue,
+            profileType: attributes.profileType?.rawValue,
```
DechengMa, line 21: For enum variable like this, it will be alright for fetching from the server and convert to string. But there might need an extra reconvert if we want to read data model from local files (eg. read string data and convert back to enum for API requesting).

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
```diff
@@ -2,19 +2,14 @@
 
 import AppStoreConnect_Swift_SDK
 import Foundation
-import struct Model.App
+import Model
```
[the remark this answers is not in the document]
Nafisa2020: Probably import Module when you need to refer to multiple models but import a specific one if you need only that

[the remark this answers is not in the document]
adamjcampbell: This is an interesting one actually

Swift only has whole module imports so when we refer to the specific `struct` or other type we do it for namespacing purposes. It's not like java where you only import particular sub packages

```swift
import struct Model.App
```
Is roughly equivalent to:
```swift
import Model
typealias App = Model.App
```
It doesn't really declare a new typealias but it tells the compiler that for this file `App` refers to `Model.App`

Sources/AppStoreConnectCLI/Model/Profile.swift
adamjcampbell: Yeah there will need to be some enum handling
The plan will probably be to declare our own enums in `Model` and then to house all the existing model conversions in `Network`
Even though I don't really enjoy the duplication

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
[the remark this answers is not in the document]
Nafisa2020: As per https://thoughtbot.com/blog/swift-imports I though we can import only submodules.

Sources/AppStoreConnectCLI/Model/Build.swift
adamjcampbell: I not sure this level of consistency is worth worrying about too much
Some of these rows we do a line by line nil coalesce and others we do a `compactMap`
It's a simple computed variable that needs to return a valid row

Sources/AppStoreConnectCLI/Model/PrereleaseVersion.swift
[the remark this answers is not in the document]
adamjcampbell: 🤔 I think my explanation might not have been entirely accurate. That seems like a decent article explaining it though.
I did attend this talk a few years ago: https://www.youtube.com/watch?v=8OMQ2DiS-zA
It shows that there's no actual performance improvement

PR #164, opened 2020-05-25
Move app id / bundle id to a new common option group

### LINKED ISSUE #161: Move app id bundle id input identifier to a new Options group
# 🗣 Context

Several commands are accepting either app id or bundle id. The identifier should be moved to a new option group  to avoid duplicate declarations.

The following commands are using the identifier.
ListBetaGroupsCommand
ReadAppCommand
ListBetaTestersCommand
ListPreReleaseVersionsCommand
ReadPreReleaseVersionCommand

Nafisa2020:
Closes #161 

# 📝 Summary of Changes

- Used a common identifier for app id / bundle id 
- Added option groups  to accept app id / bundle id as 
    1.  app id or bundle id as an argument
    2.  an array of app id / bundle ids as an option

- Changed several commands accepting 
   1.  app id or bundle id as an argument
   2.  an array of app id / bundle ids as an option

## 🔨 How To Test
```swift run asc testflight apps read iba.test2```

```swift run asc testflight builds read iba.test2 10 1.0```

``` swift run asc testflight prereleaseversion list --filter-identifiers iba.test2```

```swift run asc testflight betatesters list --filter-identifiers iba.test2```

```swift run asc testflight betagroup list --filter-identifiers iba.test2```

### DIFF
--- a/Sources/AppStoreConnectCLI/Arguments/Identifier.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/Identifier.swift
@@ -1,0 +1,15 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+enum Identifier {
+     case appId(String)
+     case bundleId(String)
+
+     init(_ argument: String) {
+         switch Int(argument) == nil {
+         case true:
+             self = .bundleId(argument)
+         case false:
+             self = .appId(argument)
+         }
+     }
+ }
--- a/Sources/AppStoreConnectCLI/Arguments/IdentifierArgument.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/IdentifierArgument.swift
@@ -1,0 +1,15 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+
+struct IdentifierArgument: ParsableArguments {
+    @Argument(
+        help: ArgumentHelp(
+            "The app AppStore ID. eg. 432156789 or app bundle identifier. eg. com.example.App",
+            discussion: "Please input either app id or bundle Id",
+            valueName: "app-id / bundle-id"
+        ),
+        transform: Identifier.init
+    )
+    var identifier: Identifier
+}
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
@@ -68,6 +68,12 @@ struct ListBetaTestersCommand: CommonParsableCommand {
     @Option(help: "Number of included related resources to return.")
     var relatedResourcesLimit: Int?
 
+    func validate() throws {
+        if (!appLookupOptions.filterIdentifiers.isEmpty) && !filterGroupNames.isEmpty {
+            throw ValidationError("Only one of these relationship filters ('app-id/ bundle-id', 'group-name') can be applied.")
+        }
+    }
+
     func run() throws {
         let service = try makeService()
 
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
@@ -44,7 +44,7 @@ struct ListBetaTestersCommand: CommonParsableCommand {
     ) var filterInviteType: BetaInviteType?
 
     @OptionGroup()
-    var identifierOptions: IdentifierOptions
+    var appLookupOptions: AppLookupOptions
 
     @Option(
         parsing: .upToNextOption,
@@ -76,7 +76,7 @@ struct ListBetaTestersCommand: CommonParsableCommand {
             firstName: filterFirstName,
             lastName: filterLastName,
             inviteType: filterInviteType,
-            filterIdentifiers: identifierOptions.filterIdentifiers,
+            filterIdentifiers: appLookupOptions.filterIdentifiers,
             groupNames: filterGroupNames,
             sort: sort,
             limit: limit,
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
@@ -3,7 +3,6 @@
 import ArgumentParser
 import AppStoreConnect_Swift_SDK
 import Combine
-import Foundation
 
 struct ListBetaTestersCommand: CommonParsableCommand {
     static var configuration = CommandConfiguration(
@@ -44,23 +43,8 @@ struct ListBetaTestersCommand: CommonParsableCommand {
         )
     ) var filterInviteType: BetaInviteType?
 
-    @Option(
-        parsing: .upToNextOption,
-        help: ArgumentHelp(
-            "Filter by app AppStore ID. eg. 432156789",
-            discussion: "This option is mutually exclusive with --filter-bundle-ids.",
-            valueName: "app-id"
-        )
-    ) var filterAppIds: [String]
-
-    @Option(
-        parsing: .upToNextOption,
-        help: ArgumentHelp(
-            "Filter by app bundle identifier. eg. com.example.App",
-            discussion: "This option is mutually exclusive with --filter-app-ids.",
-            valueName: "bundle-id"
-        )
-    ) var filterBundleIds: [String]
+    @OptionGroup()
+    var identifierOptions: IdentifierOptions
 
     @Option(
         parsing: .upToNextOption,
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: Sources/AppStoreConnectCLI/Arguments/Identifier.swift [hunk 1]
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+enum Identifier {
```
DechengMa, line 3: Identifier here is not very clear
```suggestion
enum App.Identifier {
```

Sources/AppStoreConnectCLI/Arguments/IdentifierArgument.swift
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+
+struct IdentifierArgument: ParsableArguments {
```
DechengMa, line 5: Same above

Sources/AppStoreConnectCLI/Arguments/Identifier.swift
Nafisa2020: It can be an app id or a bundle id..so I am not sure whether calling it App.Identifier is justified or misleading.. should I rename it AppOrBundleIdentifier or something like that?

DechengMa: I think `App.Identifier` is all right, either bundle Id or app Id is actually the unique identifier for an App, so you can either use bundle Id or app Id to identify an app resource. Can wait for @adamjcampbell for another look on this :)

Nafisa2020: True you can use app id or bundle id to identify an app but App.Identifier may be also read as an expanded form of appID. ..just by looking at the name App.IDentifier it doesn't jump out that  it can also be a bundle id..

adamjcampbell: 🤔 I agree that `AppIdentifier` is more specific than just `Identifier`. We do run the risk of having more name clashes. The intent of this object is actually that we use it to "choose how to find/lookup an app"  

So maybe we could work along those lines, `AppLookupIdentifier` or something, still sounds odd. But it does convey intent would pair with `AppLookupOptions` which sounds okay.

Let's go `AppIdentifier` for now maybe and come back to it

Nafisa2020: AppLookupIdentifier sounds nice

Nafisa2020: I will change it to AppLookUpIdentifier

adamjcampbell: Lookup is defined as a single word so we can use the casing `AppLookupIdentifier` as opposed to `AppLookUpIdentifier` separating the words conveys a slightly different meaning

Nafisa2020: ok. Should I also change IdentifierArgument and IdentifierOptions to AppLookupArgument and AppLookupOptions?

Nafisa2020: @adamjcampbell done..please review..thx

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -174,20 +174,28 @@ class AppStoreConnectService {
...
-        if !bundleIds.isEmpty && appIds.isEmpty {
-            appIds = try GetAppsOperation(options: .init(bundleIds: bundleIds))
-                .execute(with: requestor)
-                .await()
-                .map(\.id)
+        var filterAppIds: [String] = []
+        var filterBundleIds: [String] = []
+
+        _ = filterIdentifiers.map { identifier in
+            switch (identifier) {
```
DechengMa, line 188:
```suggestion
        filterIdentifiers.forEach { identifier in
            switch identifier {
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -308,16 +316,25 @@ class AppStoreConnectService {
...
-        let operation = GetAppsOperation(options: .init(bundleIds: bundleIds))
-        let appIds = try operation.execute(with: requestor).await().map(\.id)
+    func listBetaGroups(filterIdentifiers: [AppLookupIdentifier], names: [String]) throws -> [BetaGroup] {
+        var filterAppIds: [String] = []
+        var filterBundleIds: [String] = []
 
-        return try listBetaGroups(appIds: appIds, names: names)
-    }
+        _ = filterIdentifiers.map { identifier in
+            switch (identifier) {
```
DechengMa, line 324:
```suggestion
        filterIdentifiers.forEach { identifier in
            switch identifier {
```

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
```diff
@@ -84,16 +68,6 @@ struct ListBetaTestersCommand: CommonParsableCommand {
...
 
-    func validate() throws {
-        if !filterAppIds.isEmpty && !filterBundleIds.isEmpty {
-            throw ValidationError("Filtering by both Bundle ID and App ID is not supported!")
-        }
-
-        if (!filterAppIds.isEmpty || !filterBundleIds.isEmpty) && !filterGroupNames.isEmpty {
-            throw ValidationError("Only one of these relationship filters ('app-id, bundle-id', 'group-name') can be applied.")
-        }
-    }
```
DechengMa, line 95: Is there any way that we can keep this validation? (eg. throw Error when user input both identifiers and group names)

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
Nafisa2020: Is there any extra benefit of using forEach over map?

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
Nafisa2020: The user will either enter app id or bundle id. So one id can't be both. So this validation is not required.

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
Nafisa2020: I think performance wise both are same only map is returning an array which we don't need but for each doesn't..so I will use foreach as you have suggested..let me know if there is any performance benefit .

Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift
DechengMa: there is a group name as well `!filterGroupNames.isEmpty`. I remember input both group name and appId is not supported yet by Apple store connect API

Nafisa2020: Thanks for pointing out that. So I will do something like this 
```func validate() throws {
        if (!appLookupOptions.filterIdentifiers.isEmpty) && !filterGroupNames.isEmpty {
            throw ValidationError("Only one of these relationship filters ('app-id/ bundle-id', 'group-name') can be applied.")
        }
    }```

Does this look okay to you?

Comment (DechengMa) [reply Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift:95]: yes this looks good

File: Sources/AppStoreConnectCLI/Commands/TestFlight/BetaTesters/ListBetaTestersCommand.swift [hunk 6]
```diff
@@ -85,12 +69,8 @@ struct ListBetaTestersCommand: CommonParsableCommand {
     var relatedResourcesLimit: Int?
 
     func validate() throws {
-        if !filterAppIds.isEmpty && !filterBundleIds.isEmpty {
-            throw ValidationError("Filtering by both Bundle ID and App ID is not supported!")
-        }
-
-        if (!filterAppIds.isEmpty || !filterBundleIds.isEmpty) && !filterGroupNames.isEmpty {
-            throw ValidationError("Only one of these relationship filters ('app-id, bundle-id', 'group-name') can be applied.")
+        if (!appLookupOptions.filterIdentifiers.isEmpty) && !filterGroupNames.isEmpty {
```
DechengMa, line 72:
```suggestion
        if !appLookupOptions.filterIdentifiers.isEmpty && !filterGroupNames.isEmpty {
```

DechengMa: just a minor suggestion otherwise looks good

PR #165, opened 2020-05-26
Bug/add pagination to operations

### LINKED ISSUE #92: Bug: No paging support

DechengMa:
Closes #92 

# 📝 Summary of Changes

Add pagination to 

- [x] List BundleIds
- [x] List Devices (in #181)
- [x] List Certificates
- [ ] List Profiles (not support by SDK yet)
- TestFlight
  - [x] List Apps
  - [x] List Beta Groups
  - [x] List Beta Testers
  - [x] List Builds
  - [x] List PreRelease Versions
- Users
  - [x] List Invitations (in #184)
  - [ ] List Visible Apps (may not be in 0.1.0)
  - [x] List Users 

# 🧐🗒 Reviewer Notes
Please point out if there are any operations that need pagination, if there are not included in this PR :)

### Problems:
This might break command line option:`--limit`. Can do some validation for option.limit is nil or not, then determine if pagination should be done or not.

## 🔨 How To Test

`swift run asc certificates list `
`swift run asc testflight prereleaseversion list`
`swift run asc users list`

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
@@ -23,9 +23,9 @@ struct ListAppsOperation: APIOperation {
     func execute(with requestor: EndpointRequestor) -> AnyPublisher<[App], Error> {
         var filters: [ListApps.Filter] = []
 
-        options.bundleIds.isEmpty ? () : filters.append(.bundleId(options.bundleIds))
-        options.names.isEmpty ? () : filters.append(.name(options.names))
-        options.skus.isEmpty ? () : filters.append(.sku(options.skus))
+        if options.bundleIds.isNotEmpty { filters.append(.bundleId(options.bundleIds)) }
+        if options.names.isNotEmpty { filters.append(.name(options.names)) }
+        if options.skus.isNotEmpty { filters.append(.sku(options.skus)) }
 
         let limits = options.limit.map { [ListApps.Limit.apps($0)] }
 
--- a/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
@@ -10,17 +10,6 @@ struct ListAppsOperation: APIOperation {
         let names: [String]
         let skus: [String]
         let limit: Int?
-
-        fileprivate var endpoint: APIEndpoint<AppsResponse> {
-            var filters: [ListApps.Filter] = []
-            bundleIds.isEmpty ? () : filters.append(.bundleId(bundleIds))
-            names.isEmpty ? () : filters.append(.name(names))
-            skus.isEmpty ? () : filters.append(.sku(skus))
-
-            let limits = limit.map { [ListApps.Limit.apps($0)] }
-
-            return .apps(filters: filters, limits: limits)
-        }
     }
 
     private let options: Options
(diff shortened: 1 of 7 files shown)


orj:
@DechengMa I think the full list of paginated responses is:
- List BundleIds
- List Devices
- List Certificates (done in this PR)
- List Profiles
- TestFlight
  - List Apps
  - List Beta Groups
  - List Beta Testers
  - List Builds (done previously?)
  - List PreRelease Versions (done in this PR)
- Users
  - List Invitations
  - List Visible Apps
  - List Users (done in this PR)

orj:
@DechengMa please make sure pagination is applied to all the places it makes sense before closing #92

Sources/AppStoreConnectCLI/Services/Operations/ListCertificatesOpertaion.swift
```diff
@@ -18,11 +20,7 @@ struct ListCertificatesOperation: APIOperation {
         }
     }
 
-    private let endpoint: APIEndpoint<CertificatesResponse>
-
-    init(options: ListCertificatesOptions) {
-        typealias Filter = Certificates.Filter
-
+    var filters: [Certificates.Filter] {
```
orj, line 23:
```suggestion
    var filters: [Filter] {
```

Sources/AppStoreConnectCLI/Services/Operations/ListUsersOperation.swift
```diff
@@ -6,42 +6,51 @@ import struct Model.User
...
 
-    init(options: ListUsersOptions) {
-        let include = options.includeVisibleApps ? [ListUsers.Include.visibleApps] : nil
-
-        let limit = [
-            options.limitUsers.map(ListUsers.Limit.users),
-            options.limitVisibleApps.map(ListUsers.Limit.visibleApps)]
-            .compactMap { $0 }
-            .nilIfEmpty()
+    var limit: [ListUsers.Limit]? {
```
orj, line 12: Why not use the typealias?
```suggestion
    var limit: [Limit]? {
```

Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
```diff
@@ -32,7 +21,27 @@ struct ListAppsOperation: APIOperation {
     typealias App = AppStoreConnect_Swift_SDK.App
 
     func execute(with requestor: EndpointRequestor) -> AnyPublisher<[App], Error> {
-        requestor.request(options.endpoint).map(\.data).eraseToAnyPublisher()
+        var filters: [ListApps.Filter] = []
+
+        options.bundleIds.isEmpty ? () : filters.append(.bundleId(options.bundleIds))
```
orj, line 26: 🤔 Ok, so this is cute.  But I'm not sure I'm a fan of it.  Generally I'd suggest that ternary operators shouldn't be used for side effects.

```suggestion
        if options.bundleIds.isNotEmpty { filters.append(.bundleId(options.bundleIds)) }
```
This would require:
```
extension Collection {
    var isNotEmpty: Bool { isEmpty == false }
}
``` 
to be defined somewhere in the project. Or alternatively:
```suggestion
        if !options.bundleIds.isEmpty { filters.append(.bundleId(options.bundleIds)) }
```

But we tend to want to avoid the `!` not operator.

Sources/AppStoreConnectCLI/Services/Operations/ListBundleIdsOperation.swift
```diff
@@ -33,6 +21,24 @@ struct ListBundleIdsOperation: APIOperation {
     typealias BundleId = AppStoreConnect_Swift_SDK.BundleId
 
     func execute(with requestor: EndpointRequestor) throws -> AnyPublisher<[BundleId], Error> {
-        requestor.request(options.endpoint).map(\.data).eraseToAnyPublisher()
+        let platforms = options.platforms.compactMap(Platform.init(rawValue:))
+
+        var filters: [BundleIds.Filter] = []
+        filters += options.identifiers.isEmpty ? [] : [.identifier(options.identifiers)]
```
orj, line 27: We should pick one way of doing this.  This is less "cute" than the alternative above. But then avoiding "cute"/"smart"/"tricky" code is preferred.

PR #170, opened 2020-05-27
Update the List Apps command to use the operation pattern

adamjcampbell:
# 📝 Summary of Changes

- Add `ListAppsOperation`
- Add `listApps` to the service

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

This a straight refactor: `asc testflight apps` should still function as expected

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
@@ -1,0 +1,38 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import AppStoreConnect_Swift_SDK
+import Combine
+
+struct ListAppsOperation: APIOperation {
+
+    struct Options {
+        let bundleIds: [String]
+        let names: [String]
+        let skus: [String]
+        let limit: Int?
+
+        fileprivate var endpoint: APIEndpoint<AppsResponse> {
+            var filters: [ListApps.Filter] = []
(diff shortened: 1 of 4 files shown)


Sources/AppStoreConnectCLI/Services/Operations/ListAppsOperation.swift
```diff
@@ -0,0 +1,38 @@
...
+        let bundleIds: [String]
+        let names: [String]
+        let skus: [String]
+        let limit: Int?
+
+        fileprivate var endpoint: APIEndpoint<AppsResponse> {
+            var filters: [ListApps.Filter] = []
+            bundleIds.isEmpty ? () : filters.append(.bundleId(bundleIds))
+            names.isEmpty ? () : filters.append(.name(names))
+            skus.isEmpty ? () : filters.append(.sku(skus))
```
adamjcampbell, line 18: This is pretty odd syntax using `()` or `Void` as a no-op
We do this kind of thing a lot but instead append an empty array e.g. `+= isEmpty ? [] : [something]`
What's our favourite version of this kind of thing?

I assume we don't want single like if statements?
`if !names.isEmpty { filters.append(.name(names)) }`
Although this ternary is effectively that so wouldn't be totally opposed

DechengMa: We use `+= []` a lot in the existing code, but happy to change coz append an empty array is a bit unnecessary.
I used `!something.isEmtpy ? array.append(filter) : nil` before. 
`if !something.isEmpty { append }` and `()` also look good to me.

PR #171, opened 2020-05-27
Update list bundle ids to use Operation

adamjcampbell:
# 📝 Summary of Changes

- Adds `ListBundleIdsOperation`
- Adds `listBundleIds` service function

# 🧐🗒 Reviewer Notes
## 🔨 How To Test

This is a straight refactor `asc bundleids list` usage should remain unchanged

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/BundleIds/ListBundleIdsCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/BundleIds/ListBundleIdsCommand.swift
@@ -2,8 +2,6 @@
 
 import AppStoreConnect_Swift_SDK
 import ArgumentParser
-import Foundation
-import struct Model.BundleId
 
 struct ListBundleIdsCommand: CommonParsableCommand {
     public static var configuration = CommandConfiguration(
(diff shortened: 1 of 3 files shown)


Sources/AppStoreConnectCLI/Commands/BundleIds/ListBundleIdsCommand.swift
```diff
@@ -17,64 +15,32 @@ struct ListBundleIdsCommand: CommonParsableCommand {
...
-            .await()
-
-        bundleId.render(format: common.outputFormat)
-    }
-}
+    @Option(
+        parsing: .upToNextOption,
+        help: "Filter the results by platform (\(Platform.allCases.description))."
+    )
+    var filterPlatform: [String]
```
DechengMa, line 28: Do we want to have `[String]` here and move the `enum Platform` validation into the operation?

adamjcampbell: At the moment there's a conversion and incorrect params are ignored. We might want to have a `Platform` type sitting in `Model` that we convert from when we call the API. Even though that's some duplication :/

PR #175, opened 2020-05-28
Filter betatesters by betagroups

Nafisa2020:
# 📝 Summary of Changes

-  Filter beta testers by beta groups

# 🧐🗒 Reviewer Notes
- Betatesters API doesn't support filtering by both identifier and beta groups. When  we need to filter by both identifier and beta group names we need to use the  betagroup endpoint with included option .app and .betatesters.

## 🔨 How To Test

```swift run asc testflight betatesters list --filter-identifiers iba.test3 --filter-group-names testGrp1```


adamjcampbell:
Is this PR making use of the endpoint described here: https://developer.apple.com/documentation/appstoreconnectapi/list_all_beta_testers_in_a_betagroup ?

I feel like there's room for a different `Operation` making use of this endpoint to make some of this code easier to reason about

Nafisa2020:
@adamjcampbell When group name and identifiers are both passed I am calling the betagroup endpoint with included beta testers

Nafisa2020:
This is no longer required as we have #176

PR #206, opened 2020-07-22
Implement create profile command 

### LINKED ISSUE #200: Bug: `asc profiles create` subcommand is not implemented and causes fatal error
# 🌎 Environment

 - OS: macOS
 - Version/Release: 10.15.5

# 💬 Description

Running the `profiles` subcommand `create` causes a fatal error.

# 🦶 Reproduction Steps

Steps to reproduce the behavior, provide an example of the command line issued:
1. From master, run this `asc profiles create test ios_app_store com.fake.example`

# 🤔 Expected Results

The command should return an error message from the API about the bundle id not existing

# 😲 Actual Results

```
Fatal error: Not implementedf: file /private/var/folders/7t/2bhgs_c9099frbz84gdf54600000gn/T/mint/github.com_ittybittyapps_appstoreconnect-cli/Sources/AppStoreConnectCLI/Commands/Profiles/CreateProfileCommand.swift, line 31
```

## 🌳 Logs

```
N/A
```

## 📄 Stack Traces

```
N/A
```

# 🤝 Relationships

- Related PRs???

DechengMa:
Closes #200 

# 📝 Summary of Changes

- Create `CreateProfileOperation` 
- Implement `CreateProfileCommand`

# 🧐🗒 Reviewer Notes

## 💁 Example

Usage
```
asc profiles create <name> <profile-type> <bundle-id> [--certificates <certificates> ...] [--devices <devices> ...]
```

## 🔨 How To Test

```
asc profiles create fooname ios_app_development com.example.foo 1234abcd 1234abcd
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
+++ b/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
@@ -719,22 +719,19 @@ class AppStoreConnectService {
         .await()
         .id
 
-        let deviceIds = try deviceUDIDs.compactMap {
-            try ListDevicesOperation(
-                options: .init(
-                    filterName: [],
-                    filterPlatform: [],
-                    filterUDID: [$0],
-                    filterStatus: nil,
-                    sort: nil,
-                    limit: nil
-                )
+        let deviceIds = try ListDevicesOperation(
+            options: .init(
+                filterName: [],
+                filterPlatform: [],
+                filterUDID: deviceUDIDs,
+                filterStatus: nil,
+                sort: nil,
+                limit: nil
             )
-            .execute(with: requestor)
-            .await()
-            .first?
-            .id
-        }
+        )
+        .execute(with: requestor)
+        .await()
+        .map { $0.id }
 
         let certificateIds = try certificateSerialNumbers.compactMap {
             try ListCertificatesOperation(
--- a/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
+++ b/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
@@ -95,6 +95,7 @@ class AppStoreConnectService {
             )
             .execute(with: requestor)
             .await()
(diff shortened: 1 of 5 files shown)


DechengMa:
@orj I did some tweaks and for not exposing internal Id to the user, the input was changed to 
```
var certificatesSerialNumbers: [String]

var devicesUdids: [String]
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -704,6 +705,67 @@ class AppStoreConnectService {
...
+        deviceUDIDs: [String]
+    ) throws -> Model.Profile {
+        let bundleIdResourceId = try ReadBundleIdOperation(
+            options: .init(bundleId: bundleId)
+        )
+        .execute(with: requestor)
+        .await()
+        .id
+
+        let deviceIds = try deviceUDIDs.compactMap {
```
orj, line 722: This maybe could be a single call rather than multiple.  You can filter devices by multiple UDID I think.

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -704,6 +705,67 @@ class AppStoreConnectService {
...
+            .execute(with: requestor)
+            .await()
+            .first?
+            .id
+        }
+
+        let certificateIds = try certificateSerialNumbers.compactMap {
+            try ListCertificatesOperation(
+                options: .init(
+                    filterSerial: $0,
```
orj, line 742: Likewise I think this could be a single call rather than multiple. You should be able to filter by multiple serials.

DechengMa: Will skip this for now as the current `ListCertificatesOperation` doesn't support passing an array of `filterSerials`.
Don't want to make this PR too massive, If there is a problem with requesting time, I will create another PR for fixing this entirely.

PR #220, opened 2020-08-17
Adding a verbose flag for displaying more messages in terminal

csjones:
# 🤔 Objective and Motivation

The objective of these changes are to purpose adding an optional flag `-v/--verbose` to display more output in terminal. My motivation for this flag is to remove overhead when using `asc` with ansible. I'm using the following command in my ansible playbook:

```shell
asc profiles list --json --filter-name "App Store {{ app_name }}" --download-path .
```
...and I get the following output:
```
"stdout": "📥 Profile 'App Store Test001010002' downloaded to: ./aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa.mobileprovision\n[{\"createdDate\":619333688,\"expirationDate\":620178602,\"name\":\"App Store Test001010002\",\"platform\":\"IOS\",\"profileContent\":\"MdMIIdqAAYJKoZIhvcNAQQcCoIIdmTCCHZCUCAQExCzACJB........
```

The `"stdout"` is now malformed json and requires additional parsing to work correctly by filtering out the `"📥 Profile 'App Store Test001010002' downloaded to: ./aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa.mobileprovision\n"`. With the `--verbose` flag added, the command now looks like this:
```shell
asc profiles list --verbose --json --filter-name "App Store {{ app_name }}" --download-path .
```
...and I get the following output:
```
"stdout": "📥 Profile 'App Store Test001010002' downloaded to: ./aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa.mobileprovision\n[{\"createdDate\":619333688,\"expirationDate\":620178602,\"name\":\"App Store Test001010002\",\"platform\":\"IOS\",\"profileContent\":\"MdMIIdqAAYJKoZIhvcNAQQcCoIIdmTCCHZCUCAQExCzACJB........
```

...and the default command and output looks like this: 

```shell
asc profiles list --json --filter-name "App Store {{ app_name }}" --download-path .
```
...and the json returned to stdout is no longer malformed so now I get the following output:
```
"stdout": "[{\"createdDate\":619333688,\"expirationDate\":620178602,\"name\":\"App Store Test001010002\",\"platform\":\"IOS\",\"profileContent\":\"MdMIIdqAAYJKoZIhvcNAQQcCoIIdmTCCHZCUCAQExCzACJB........
```
# 📝 Summary of Changes

- Adds the optional flag `-v/--verbose` to display more in terminal.
- Changed output of all commands to be parsable by default
- Removed `.prettyPrinted` and created https://github.com/ittybittyapps/appstoreconnect-cli/issues/221

# ⚠️ Items of Note

_Document anything here that you think the reviewer(s) of this PR may need to know, or would be of specific interest._

The `--verbose` only changes the output of a `CommonParsableCommand` that has not been commented with a `TODO`.

# 🧐🗒 Reviewer Notes

## 💁 Example

```
swift run asc profiles list --verbose --json --filter-name "Cool App" --download-path ~/Documents/Cool\ App
```

## 🔨 How To Test

Run these command and notice the messaging in terminal:

```
swift run asc profiles list --json --filter-name "Cool App" --download-path ~/Documents/Cool\ App
swift run asc users sync ~/Documents/fake.json --dry-run
swift run asc certificates list
swift run asc certificates read "M109UYTGSF007" --certificate-output distribution.cer
swift run asc certificates revoke "M109UYTGSF007"
```

Run the same commands with `-v/--verbose` added and notice more messaging in terminal:

```
swift run asc profiles list --verbose --json --filter-name "Cool App" --download-path ~/Documents/Cool\ App
swift run asc users sync -v ~/Documents/fake.json --dry-run
swift run asc certificates list -v
swift run asc certificates read --verbose "M109UYTGSF007" --certificate-output distribution.cer
swift run asc certificates revoke --verbose "M109UYTGSF007"
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
@@ -55,8 +55,8 @@ struct ListCertificatesCommand: CommonParsableCommand {
 
                 let file = try certificateProcessor.write($0)
 
-                // Command output is parsable by default. Only print if user is enabling verbosity or output is a `.table`
-                if common.verbose || common.outputFormat == .table {
+                // Only print if the `PrintLevel` is set to verbose.
+                if common.printLevel == .verbose {
                     print("📥 Certificate '\($0.name ?? "")' downloaded to: \(file.path)")
                 }
             }
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
@@ -55,7 +55,8 @@ struct ListCertificatesCommand: CommonParsableCommand {
 
                 let file = try certificateProcessor.write($0)
 
-                if common.quiet == false {
+                // Command output is parsable by default. Only print if user is enabling verbosity or output is a `.table`
+                if common.verbose || common.outputFormat == .table {
                     print("📥 Certificate '\($0.name ?? "")' downloaded to: \(file.path)")
                 }
             }
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
@@ -55,7 +55,9 @@ struct ListCertificatesCommand: CommonParsableCommand {
 
                 let file = try certificateProcessor.write($0)
 
-                print("📥 Certificate '\($0.name ?? "")' downloaded to: \(file.path)")
+                if common.quiet == false {
+                    print("📥 Certificate '\($0.name ?? "")' downloaded to: \(file.path)")
+                }
             }
         }
 
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/ReadCertificateCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/ReadCertificateCommand.swift
@@ -28,8 +28,8 @@ struct ReadCertificateCommand: CommonParsableCommand {
 
             let file = try fileProcessor.write(certificate)
 
-            // Command output is parsable by default. Only print if user is enabling verbosity or output is a `.table`
-            if common.verbose || common.outputFormat == .table {
+            // Only print if the `PrintLevel` is set to verbose.
+            if common.printLevel == .verbose {
                 print("📥 Certificate '\(certificate.name ?? "")' downloaded to: \(file.path)")
             }
         }
--- a/Sources/AppStoreConnectCLI/Commands/Certificates/ReadCertificateCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/Certificates/ReadCertificateCommand.swift
@@ -28,7 +28,9 @@ struct ReadCertificateCommand: CommonParsableCommand {
 
             let file = try fileProcessor.write(certificate)
 
-            print("📥 Certificate '\(certificate.name ?? "")' downloaded to: \(file.path)")
+            if common.quiet == false {
+                print("📥 Certificate '\(certificate.name ?? "")' downloaded to: \(file.path)")
+            }
         }
 
         certificate.render(format: common.outputFormat)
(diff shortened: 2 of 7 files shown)


adamjcampbell:
Hey @csjones, thank you for your contribution. You've prompted a discussion on our end with this PR about whether quiet, output only commands should be the default response. That is, whether or not this should be implemented as only logging these `Certificate downloaded` `Certificate revoked` messages when a `--verbose` flag is passed.

This would then need to be rolled out across other commands so that the output of all commands is parsable by default. Did you have an extra thoughts on the matter for your use case?

orj:
I agree with @adamjcampbell here.  I think the stuff like `📥 Profile 'X' downloaded to: ...` should only be printed when a `--verbose` flag was in effect.

🤔 One thing to consider is perhaps when using `--json`, `--yaml`, and `--csv`, `--verbose` could be `false` but with `--table` it could be `true`?  Given that the computer parsable formats are more likely to be used in pipes than the table output which is designed for interactive use.  Thoughts?

csjones:
Hey @adamjcampbell @orj, I'm in favor of these changes. I think machine parsable by default and a `--verbose` option instead of `--quiet` makes a lot sense, especially for my use case. Also makes sense to default `--verbose` to `true` when using the human parseable `--table` option.

I would propose one more additional change. When using `--json` (or any machine parsable format) to minimize the output by default instead of pretty printing the output. Currently, `--json` is using a prettify option and adds additional whitespace and newlines which needs to be filtered out.

adamjcampbell:
@csjones sounds good to me. Are you happy to make those changes for this command / this PR?

In regards to the pretty formatting of json I would think filtering out the white space would be optional. It shouldn't affect the ability to parse the JSON. However we could have it as an option or the default when not in verbose mode.

What are your thoughts on the matter?

orj:
@adamjcampbell Perhaps we can have a `--pretty` option for prettifying the `json`.  Mostly I think you would be piping json output into a tool like `jq`and `jq` can pretty it for you.

csjones:
@adamjcampbell Sure! I'll make the changes and update the PR. For pretty printing output, I'm leaning towards removing `.prettyPrinted` since `--pretty` currently does not exist and I prefer @orj's idea of using `jq` if needed because my use case is aimed towards better accommodating machine parsable output. Additionally, I could also create a new GitHub issue to add a `--pretty` option and see if any other `asc` users would prefer to have this feature flag added.

 Does this list of changes look good?

**Summary of changes:**
 - [ ] Change `--quiet` to be `--verbose`
 - [ ] `--verbose` defaults to `false` except when `--table` is used
 - [ ] Remove use of `.prettyPrinted`
 - [ ] Create GitHub issue for `--pretty` feature flag

adamjcampbell:
@csjones sounds like a plan! Thanks for all your efforts on this.

Sources/AppStoreConnectCLI/Commands/Certificates/ListCertificatesCommand.swift
```diff
@@ -55,7 +55,10 @@ struct ListCertificatesCommand: CommonParsableCommand {
 
                 let file = try certificateProcessor.write($0)
 
-                print("📥 Certificate '\($0.name ?? "")' downloaded to: \(file.path)")
+                // Command output is parsable by default. Only print if user is enabling verbosity or output is a `.table`
+                if common.verbose || common.outputFormat == .table {
```
adamjcampbell, line 59: We seem to have to use `common.verbose || common.outputFormat == .table` quite a bit here.
What do you think about making `verbose` be `Bool?` (defaulting to `nil`) and then we can add a computed var either a Bool or enum to check.

Thinking along these lines:

```swift
// CommonParsableCommand.swift

enum PrintLevel {
  case quiet
  case verbose
}

// CommonOptions

var printLevel: PrintLevel {
  switch verbose {
    case true:
      return .verbose
    case false:
      return .quiet
    case nil:
      return common.outputFormat == .table ? .verbose : .quiet
  }
}

// Usage

if common.printLevel == .verbose {
  ...
}
```

csjones: Overall, this is a good suggestion and I'll make these changes with small tweaks. I'll leave `verbose` as a `Bool` and I'll resolve the `common.outputFormat == .table` ternary in the `switch` statement itself. Should look something like this:
```swift
// CommonOptions

var printLevel: PrintLevel {
  switch (verbose, outputFormat == .table) {
    case (true, _), (_, true):
      return .verbose
    case (false, false):
      return .quiet
  }
}
```

adamjcampbell: Sounds good, thanks @csjones

adamjcampbell (approved):
@csjones This is looking ✅, just needs the swiftlint warnings attended to

PR #223, opened 2020-08-26
List, read, create, delete, update Beta Build Localizations

### LINKED ISSUE #222: List, Read, Create, Modify, Delete Beta Build Localizations
# 🗣 Context

Usually after uploading a build using `altool` you want to be able to update the what's new (changeling) details of a build. Having this capability built into `asc` would be very useful.

# 💬 Narrative

As a user
I want to be able to List, Read, Create, Modify, & Delete beta build localisations
So that I can see and update the `whatsNew` details of builds I've uploaded

# 📝 Notes

- I think you can only update the localisation details of a build once it has begun processing.
- This feature covers these [APIs](https://developer.apple.com/documentation/appstoreconnectapi/prerelease_versions_and_beta_testers/beta_build_localizations)

# 🎨 Design

```
asc testflight builds localization {list,create,read,update,delete} <bundleId> <preReleaseVersion> <buildNumber> <locale> <changelog>
```

The `changelog` could be read in from stdin (default) or perhaps specified via a file (`-i changelog.txt`).

# ✅ Acceptance Criteria

**GIVEN** there is a localisation for the specified build, preReleaseVersion & locale
**WHEN** I request that localisation
**THEN** the localisation information is displayed

**GIVEN** there there are localisation for the specified build & preReleaseVersion
**WHEN** I request a list of localisations for that build
**THEN** a list localisations is displayed

**GIVEN** I do not specify a changelog on the command line
**WHEN** I create a localisation
**THEN** the changelog is read from standard input

DechengMa:
Closes #222
- [x] List
- [x] Read
- [x] Create
- [x] Delete
- [x] Update
- [x] Read from stdin 
# 📝 Summary of Changes

- Implement List, Read, create, delete, update Beta Build localization information.

# 🧐🗒 Reviewer Notes

## 💁 Example

USAGE: asc testflight builds localization <subcommand>

```
SUBCOMMANDS:
  create                  Create localized What’s New text for a build.
  delete                  Delete a specific beta build localization associated with a build.
  list                    Find and list beta build localization resources.
  read                    Get a specific beta build localization resource.
  update                  Update the localized What’s New text for a specific beta build and locale.
```

##### Create 
```
USAGE: asc testflight builds localization create <bundle-id> <pre-release-version> <build-number> <locale> [--whats-new <whats-new>] [--path <path>]
```
##### Update
```
USAGE: asc testflight builds localization create <bundle-id> <pre-release-version> <build-number> <locale> [--whats-new <whats-new>] [--path <path>]
```

##### Delete
```
USAGE: asc testflight builds localization delete <bundle-id> <pre-release-version> <build-number> <locale>
```

##### List
```
USAGE: asc testflight builds localization list <bundle-id> <pre-release-version> <build-number> [--limit <limit>]
```

##### Read
```
USAGE: asc testflight builds localization read [--api-issuer <uuid>] [--api-key-id <keyid>] [--csv] [--json] [--table] [--yaml] <bundle-id> <pre-release-version> <build-number> <locale>
```

## 🔨 How To Test

`[--whats-new <whats-new>]` must be wrapped inside a `' '`
###### List
```
swift run asc testflight builds localization update com.example.foo 1.0.9 5
```

###### Read
```
swift run asc testflight builds localization update com.example.foo 1.0.9 5 en-AU 
```

###### Delete
```
swift run asc testflight builds localization update com.example.foo 1.0.9 5 en-AU 
```

###### Create
```
swift run asc testflight builds localization update com.example.foo 1.0.9 5 en-AU --json --path ../whatsNew.text
```

###### Update
```
swift run asc testflight builds localization update com.example.foo 1.0.9 5 en-AU --json --whats-new `Hello There`
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -34,14 +34,14 @@ extension CreateUpdateBuildLocalizationCommand {
         }
     }
 
-    func readWhatsNew() -> String {
+    func readWhatsNew() throws -> String {
         let whatsNew: String
 
         if let whatsNewFromOption = localization.whatsNew {
             whatsNew = whatsNewFromOption
         } else if let filePath = localization.path {
             guard let whatsNewString = try? String(contentsOfFile: filePath) else {
-                fatalError("Unable to read text from file path")
+                throw Error.invalidFilePath(filePath)
             }
 
             whatsNew = whatsNewString
@@ -58,4 +58,15 @@ extension CreateUpdateBuildLocalizationCommand {
         return whatsNew
     }
 
+}
+
+private enum Error: LocalizedError {
+    case invalidFilePath(String)
+
+    var errorDescription: String? {
+        switch self {
+        case .invalidFilePath(let filePath):
+            return "Unable to read text from file path: '\(filePath)'."
+        }
+    }
 }
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -34,7 +34,7 @@ extension CreateUpdateBuildLocalizationCommand {
         }
     }
 
-    var whatsNew: String {
+    func readWhatsNew() -> String {
         let whatsNew: String
 
         if let whatsNewFromOption = localization.whatsNew {
@@ -46,7 +46,6 @@ extension CreateUpdateBuildLocalizationCommand {
 
             whatsNew = whatsNewString
         } else {
-            print("Reading text from stdin. (press CTRL+D to finish)")
             var whatsNewStdin: [String] = []
 
             while let line = readLine() {
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -1,0 +1,50 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import FileSystem
+
+struct BuildLocalizationInputArguments: ParsableArguments {
+    @Argument(help: "The locale information of the build localization resource. eg. (en-AU)")
+    var locale: String
+
+    @Option(
+        parsing: SingleValueParsingStrategy.scanningForValue,
+        help: """
+            A field that describes changes and additions to a build and indicates features you would like your users to test.
+            Please put the text inside an ''.
+            eg. ('Hello There')
+        """
+    )
+    var whatsNew: String?
+
+    @Option(help: "File path to a txt file that contains whats new text. (eg. whatsNew.txt)")
+    var path: String?
+}
+
+protocol CreateUpdateBuildLocalizationCommand {
+    var localization: BuildLocalizationInputArguments { get }
+}
+
+extension CreateUpdateBuildLocalizationCommand {
+
+    func validateWhatsNewInput() throws {
+        if localization.whatsNew == nil && localization.path == nil {
+            throw ValidationError("Please either use --whatsNew or --path option to provide whats new info.")
+        }
+
+        if localization.whatsNew != nil && localization.path != nil {
+            throw ValidationError("Please provide either a file path or whats new text in terminal.")
+        }
+    }
+
+    var whatsNew: String {
+        if let whatsNew = localization.whatsNew {
+            return whatsNew
+        } else if let filePath = localization.path {
+            return Readers.FileReader<String>(format: .txt).readTXT(from: filePath)
+        }
+
+        return ""
+    }
+
+}
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -2,6 +2,7 @@
 
 import ArgumentParser
 import FileSystem
+import Foundation
 
 struct BuildLocalizationInputArguments: ParsableArguments {
     @Argument(help: "The locale information of the build localization resource. eg. (en-AU)")
@@ -39,7 +40,11 @@ extension CreateUpdateBuildLocalizationCommand {
         if let whatsNewFromOption = localization.whatsNew {
             whatsNew = whatsNewFromOption
         } else if let filePath = localization.path {
-            whatsNew = Readers.FileReader<String>(format: .txt).readTXT(from: filePath)
+            guard let whatsNewString = try? String(contentsOfFile: filePath) else {
+                fatalError("Unable to read text from file path")
+            }
+
+            whatsNew = whatsNewString
         } else {
             print("Reading text from stdin. (press CTRL+D to finish)")
             var whatsNewStdin: [String] = []
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -40,11 +40,7 @@ extension CreateUpdateBuildLocalizationCommand {
         if let whatsNewFromOption = localization.whatsNew {
             whatsNew = whatsNewFromOption
         } else if let filePath = localization.path {
-            guard let whatsNewString = try? String(contentsOfFile: filePath) else {
-                throw Error.invalidFilePath(filePath)
-            }
-
-            whatsNew = whatsNewString
+            whatsNew = try String(contentsOfFile: filePath)
         } else {
             var whatsNewStdin: [String] = []
 
@@ -58,15 +54,4 @@ extension CreateUpdateBuildLocalizationCommand {
         return whatsNew
     }
 
-}
-
-private enum Error: LocalizedError {
-    case invalidFilePath(String)
-
-    var errorDescription: String? {
-        switch self {
-        case .invalidFilePath(let filePath):
-            return "Unable to read text from file path: '\(filePath)'."
-        }
-    }
 }
--- a/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
+++ b/Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
@@ -28,23 +28,30 @@ protocol CreateUpdateBuildLocalizationCommand {
 extension CreateUpdateBuildLocalizationCommand {
 
     func validateWhatsNewInput() throws {
-        if localization.whatsNew == nil && localization.path == nil {
-            throw ValidationError("Please either use --whatsNew or --path option to provide whats new info.")
-        }
-
         if localization.whatsNew != nil && localization.path != nil {
             throw ValidationError("Please provide either a file path or whats new text in terminal.")
         }
     }
 
     var whatsNew: String {
-        if let whatsNew = localization.whatsNew {
-            return whatsNew
+        let whatsNew: String
+
+        if let whatsNewFromOption = localization.whatsNew {
+            whatsNew = whatsNewFromOption
         } else if let filePath = localization.path {
-            return Readers.FileReader<String>(format: .txt).readTXT(from: filePath)
+            whatsNew = Readers.FileReader<String>(format: .txt).readTXT(from: filePath)
+        } else {
+            print("Reading text from stdin. (press CTRL+D to finish)")
+            var whatsNewStdin: [String] = []
+
+            while let line = readLine() {
+                whatsNewStdin.append(line)
+            }
+
+            whatsNew = whatsNewStdin.joined(separator: "\n")
         }
 
-        return ""
+        return whatsNew
     }
 
 }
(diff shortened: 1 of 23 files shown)


Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/Localizations/CreateBuildLocalizationsCommand.swift
```diff
@@ -0,0 +1,39 @@
...
+        let buildLocalization = try service.createBuildLocalization(
+            bundleId: build.bundleId,
+            buildNumber: build.buildNumber,
+            preReleaseVersion: build.preReleaseVersion,
+            locale: localization.locale,
+            whatsNew: whatsNew
+        )
+
+        // Render long text by SwiftyTable is not supported yet
+        [buildLocalization].render(format: common.outputFormat == .table ? .json : common.outputFormat)
```
adamjcampbell, line 37: 🤔 should we throw an error that table isn't supported instead of rendering in a different format?

adamjcampbell: Potentially as part of the validation step now that I'm thinking on it

Sources/AppStoreConnectCLI/Model/BuildLocalization.swift
```diff
@@ -0,0 +1,32 @@
...
+            TextTableColumn(header: "Locale"),
+            TextTableColumn(header: "What's New"),
+        ]
+    }
+
+    var tableRow: [CustomStringConvertible] {
+        [
+            locale ?? "",
+            "Render whatsNew info in table is currently not support, please use --json or --yaml instead",
+        ]
```
adamjcampbell, line 29: Might just be able to `fatalError("unimplemented")` here. Since we are trying to ensure it's not being called

Sources/FileSystem/Readers.swift
```diff
@@ -67,6 +69,14 @@ public enum Readers {
 
             return result
         }
+
+        public func readTXT(from filePath: String) -> T {
```
adamjcampbell, line 73: What values of `T` are you expecting? Will it ever be something different to `String`

DechengMa: I was trying to do something like `public func readTXT<T: String>(from filePath: String) -> T {` but string can't confirm to `Decodable` so I removed it.
Do you have some thoughts on this?

Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/Localizations/CreateBuildLocalizationsCommand.swift
DechengMa: As `table` is the default format, it will be a bit user-unfriendly if we just throw error. 
What do you think of not using CommonOptions in this command, and extract these out 
```
 @OptionGroup()
var authOptions: AuthOptions

@Flag(default: .json, help: "Display results in specified format.")
var outputFormat: OutputFormat
```

Sources/FileSystem/Readers.swift
adamjcampbell: There's always overloads. If we only have need for strings atm I'd define this one as: `public func readTXT(from filePath: String) -> String` and we can add a generic one later down the track if we need it

Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/Localizations/CreateBuildLocalizationsCommand.swift
adamjcampbell: What's the actual issue? The text is too long for the table? Maybe we can just truncate the text when rendering as a table?

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
```diff
@@ -0,0 +1,50 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import ArgumentParser
+import FileSystem
+
+struct BuildLocalizationInputArguments: ParsableArguments {
+    @Argument(help: "The locale information of the build localization resource. eg. (en-AU)")
+    var locale: String
+
+    @Option(
```
orj, line 10: When using this command I don't think you would pass the "whatsNew" content on the command line.  You would either pass it via a file or as standard input (ie, via a pipe).

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
```diff
@@ -0,0 +1,50 @@
...
+        if localization.whatsNew != nil && localization.path != nil {
+            throw ValidationError("Please provide either a file path or whats new text in terminal.")
+        }
+    }
+
+    var whatsNew: String {
+        if let whatsNew = localization.whatsNew {
+            return whatsNew
+        } else if let filePath = localization.path {
+            return Readers.FileReader<String>(format: .txt).readTXT(from: filePath)
```
orj, line 44: Really not sure we need a "text" file reader when `let text2 = try String(contentsOf: fileURL, encoding: .utf8)` is sufficient.  Reading from standard input is a bit more fiddly but still quite simple.

Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/Localizations/CreateBuildLocalizationsCommand.swift
DechengMa: > What's the actual issue? The text is too long for the table? Maybe we can just truncate the text when rendering as a table?

The table doesn't support line break and long text will break the table structure.

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
DechengMa: > When using this command I don't think you would pass the "whatsNew" content on the command line. You would either pass it via a file or as standard input (ie, via a pipe).

Hi @orj The command provides two options, you can either input via command line or file path. It might be useful if the input text is short. So do you want me to remove `--whatsNew` option?

Sources/AppStoreConnectCLI/Commands/TestFlight/Builds/Localizations/CreateBuildLocalizationsCommand.swift
adamjcampbell: The table is just meant to be a quick reference. To do any real automation/work we use an encodable format. I think we should be able to filter out the line breaks (replace with space) and truncate in this use case. We can probably alter the title of What's New the table to explain it's not the verbatim what's new.

Something like `What's New (truncated)`

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
adamjcampbell: @DechengMa I believe the suggestion is to remove the `--whatsNew` argument and if there is no file specified to read from `stdin`

orj: For a very short (one line) what's new you might have a command line option for it (eg, git has `-m "Commit message"` for commits.  I guess we could keep it for those cases.  But the "unix philosophy" would be to take the input from standard input usually:
```
asc testflight builds localization create com.bundleid 1.0 100 en_AU < changelog.txt
```
You could also do one-liners with:
```
echo "Bugfixes and performance improvements" | asc testflight builds localization create com.bundleid 1.0 100 en_AU
```

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
DechengMa: I've tried `String(contentsOf: )` but appears needs to force wrap the string to URL twice.
```
guard let fileURL = URL(string: filePath),
      let whatsNewString = try? String(contentsOf: fileURL, encoding: .utf8) 
else { }
``` 
I think put it in Reader can let use do error handling centrally. 🤔

Sources/FileSystem/Readers.swift
DechengMa: > There's always overloads. If we only have need for strings atm I'd define this one as: `public func readTXT(from filePath: String) -> String` and we can add a generic one later down the track if we need it

As `FileReader<T: Decodable>` has generic T:Decodable type so this odd syntax is to conform to it. Otherwise, it will give me `Cannot convert return expression of type 'String' to return type 'T'` 🤔 

If this really looks bad how about we make it as `static`?

adamjcampbell: If you are taking @orj 's suggestion to make use of `String(contentsOf:encoding:)` we no longer need this function

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
DechengMa: This was implemented. Now these options are provided 

```
--whats-new "Some short text"
```

Or 
```
 --path ../whatsNew.text
```

Or `stdin`
```
echo "How about test this" | swift run asc testflight builds localization update com.exmaple.foo 1.0.9 5 en-AU   
Reading text from stdin. (press CTRL+D to finish)
+---------+------------------------+
| Locale  | What's New (truncated) |
+---------+------------------------+
|  en-AU  | "How about test this"  |
+---------+------------------------+
```
```
swift run asc testflight builds localization update com.exmaple.foo 1.0.9 5 en-AU                        
[3/3] Linking asc
Reading text from stdin. (press CTRL+D to finish)
Hello  
Please test these 
Cheers 
+---------+-----------------------------------+
| Locale  | What's New (truncated)            |
+---------+-----------------------------------+
| en-AU   | "Hello Please test these  Cheers" |
+---------+-----------------------------------+
```

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    var whatsNew: String {
+        let whatsNew: String
+
+        if let whatsNewFromOption = localization.whatsNew {
+            whatsNew = whatsNewFromOption
+        } else if let filePath = localization.path {
+            guard let whatsNewString = try? String(contentsOfFile: filePath) else {
+                fatalError("Unable to read text from file path")
```
adamjcampbell, line 44: Should we just throw the error and make getting what's new a function?

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            whatsNew = whatsNewString
+        } else {
+            print("Reading text from stdin. (press CTRL+D to finish)")
```
adamjcampbell, line 49: This is pretty common practice and doesn't need a log statement. However we should document that we read from `stdin` when a file or what's new isn't specified in the help text

DechengMa: @adamjcampbell is abstract a good place for documenting this?
```
abstract: "Create localized What’s New text for a build. Text from `stdin` will be read when a file or what's new isn't specified."
```

adamjcampbell: Maybe `discussion` is more suited in that same constructor. Reading the docs `abstract` is meant to be the one line description

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
adamjcampbell: Were you going to make `readWhatNew` a throwing function?

DechengMa: @adamjcampbell Yes, I just changed the function to a throwing function. `throw Error.invalidFilePath(filePath)`

Sources/AppStoreConnectCLI/Arguments/BuildLocalizationInputArguments.swift
```diff
@@ -0,0 +1,72 @@
...
+        let whatsNew: String
+
+        if let whatsNewFromOption = localization.whatsNew {
+            whatsNew = whatsNewFromOption
+        } else if let filePath = localization.path {
+            guard let whatsNewString = try? String(contentsOfFile: filePath) else {
+                throw Error.invalidFilePath(filePath)
+            }
+
+            whatsNew = whatsNewString
```
adamjcampbell, line 47: It's probably more idiomatic to use do catch here if we're wrapping the error:
```suggestion
            do {
                whatsNew = try String(contentsOfFile: filePath)
            } catch {
                throw Error.invalidFilePath(filePath)
            }
```

Alternatively if the errors that `String(contentsOfFile:)` throws seems descriptive enough you can lean on that error:
```suggestion
            whatsNew = try String(contentsOfFile: filePath)
```

DechengMa: the second one looks good, no need for an extra enum Error, I've changed to that, please take a look 🙂

adamjcampbell (approved):

PR #224, opened 2020-08-26
Add BetaGroup and BetaTester Model in FileSystem

DechengMa:
This PR is to create Model that represents the file on the disk for sync command  

# 📝 Summary of Changes

- Create Model BetaGroup and BetaTester in FileSystem

# 🧐🗒 Reviewer Notes
Didn't create `App` model as we have one in `Model` which can be used directly

### DIFF
--- a/Sources/FileSystem/Model/BetaGroup.swift
+++ b/Sources/FileSystem/Model/BetaGroup.swift
@@ -1,0 +1,40 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+public struct BetaGroup: Codable, Equatable {
+
+    public typealias EmailAddress = String
+
+    public var id: String?
+    public var groupName: String
+    public var isInternal: Bool?
+    public var publicLink: String?
+    public var publicLinkEnabled: Bool?
+    public var publicLinkLimit: Int?
+    public var publicLinkLimitEnabled: Bool?
+    public var creationDate: String?
+    public var testers: [EmailAddress]
+
+    public init(
+        id: String?,
+        groupName: String,
+        isInternal: Bool?,
+        publicLink: String?,
+        publicLinkEnabled: Bool?,
+        publicLinkLimit: Int?,
+        publicLinkLimitEnabled: Bool?,
+        creationDate: String?,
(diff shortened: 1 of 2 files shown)


Sources/FileSystem/Model/BetaGroup.swift
```diff
@@ -0,0 +1,40 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+public struct BetaGroup: Codable, Equatable {
+
+    public typealias EmailAddress = String
+
+    public var id: String?
```
adamjcampbell, line 9: Is this nil because we won't have an id until we `push` it?

Sources/FileSystem/Model/BetaGroup.swift
```diff
@@ -0,0 +1,40 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+public struct BetaGroup: Codable, Equatable {
+
+    public typealias EmailAddress = String
```
adamjcampbell, line 7: Maybe we want to think about having some sort of strongly typed email in our `Model` module

Sources/FileSystem/Model/BetaGroup.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    public var id: String?
+    public var groupName: String
+    public var isInternal: Bool?
+    public var publicLink: String?
+    public var publicLinkEnabled: Bool?
+    public var publicLinkLimit: Int?
+    public var publicLinkLimitEnabled: Bool?
+    public var creationDate: String?
+    public var testers: [EmailAddress]
```
adamjcampbell, line 17: 🤔 I'm not sure any of these properties need to be mutable when we read them from or store them to disk

adamjcampbell (commented):
Looking good

Sources/FileSystem/Model/BetaGroup.swift
DechengMa: yes, when creating from dish the id will be nil

adamjcampbell (approved):

PR #226, opened 2020-08-28
Pull testflight models and pass them to the configuration processor

adamjcampbell:
# 📝 Summary of Changes

- Add `TestflightConfigurationProcessor`
- Fetch Apps, Testflight testers and groups
- Feeds models into the configuration processor
- Removes protocol requirements from `ResourceWriter`

# ⚠️ Items of Note

- This is an intentionally naive implementation that calls out the room for optimisation in future PRs
- The Command has still not been fully implemented and will still throw an error

# 🧐🗒 Reviewer Notes

## 💁 Example

```
$ asc testflight sync pull

Error: This command has not been implemented
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
@@ -3,6 +3,7 @@
 import ArgumentParser
 import Foundation
 import FileSystem
+import Model
 
 struct TestFlightPullCommand: CommonParsableCommand {
 
@@ -25,7 +26,13 @@ struct TestFlightPullCommand: CommonParsableCommand {
     ) var outputPath: String
 
     func run() throws {
-        throw CommandError.unimplemented
+        let service = try makeService()
+
+        // TODO: A new service function should be created to efficiently gather these models
+        let apps = try service.listApps(bundleIds: filterBundleIds)
+        let identifiers = apps.map { app in AppLookupIdentifier.appId(app.id) }
+        let testers = try service.listBetaTesters(filterIdentifiers: identifiers)
+        let groups = try service.listBetaGroups(filterIdentifiers: identifiers)
     }
 
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
@@ -36,6 +36,9 @@ struct TestFlightPullCommand: CommonParsableCommand {
 
         let processor = TestflightConfigurationProcessor(path: .folder(path: outputPath))
         processor.writeConfiguration(apps: apps, testers: testers, groups: groups)
+
+        // TODO: Remove when the pull is completed
+        throw CommandError.unimplemented
     }
 
 }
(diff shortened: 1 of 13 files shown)


Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
```diff
@@ -25,6 +26,18 @@ struct TestFlightPullCommand: CommonParsableCommand {
     ) var outputPath: String
 
     func run() throws {
+        let service = try makeService()
+
+        // TODO: A new service function should be created to efficiently gather these models
+        let apps = try service.listApps(bundleIds: filterBundleIds)
+        let identifiers = apps.map { app in AppLookupIdentifier.appId(app.id) }
+        let testers = try service.listBetaTesters(filterIdentifiers: identifiers)
+        let groups = try service.listBetaGroups(filterIdentifiers: identifiers)
```
DechengMa, line 35: We might need another request for requesting all the testers in the beta group

adamjcampbell: From what I can see at the moment we have all the information required.
We have all the testers for all the specified apps, we have all the groups for the specified apps and the testers list what groups they are a part of.

But maybe it will be better to do it that way when we re-write the service function...

The testers only list the group names they are a part of and they could be duplicated across apps

DechengMa: yes, that's what I am worried about, tester only has an of group names instead of `ID`.

adamjcampbell: Yeah I think you're right. I can take a look for this PR

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -0,0 +1,22 @@
...
+    public init(path: ResourcePath) {
+        self.path = path
+    }
+
+    public func writeConfiguration(
+        apps: [Model.App],
+        testers: [Model.BetaTester],
+        groups: [Model.BetaGroup]
+    ) {
+        // TODO: Convert models to a TestflightConfiguration and write it to disk
```
DechengMa, line 19: if we do the Convert in FileSystem Model. Where do we do the comparison of configs? 🤔

adamjcampbell: My feel at the moment is that we're going to convert the config into models and then we can compare the models somewhere else to produce the actions

Sources/FileSystem/Model/BetaTester.swift
```diff
@@ -2,12 +2,13 @@
...
-    public var firstName: String
-    public var lastName: String
+struct BetaTester: Codable, Hashable {
 
-    public init(
+    var email: String
+    var firstName: String
+    var lastName: String
+
+    init(
```
DechengMa, line 11: Did you also want to remove the init for BetaTester in Filesystem model as well? I saw you remove the init for BetaGroup

DechengMa (approved):
Looks good, this will be enough for us to go ahead

Sources/FileSystem/Model/BetaTester.swift
adamjcampbell: The only reason I left it in was because it was doing some work for us:  `?? ""`

PR #227, opened 2020-08-31
Write out testflight config (sync pull)

adamjcampbell:
# 📝 Summary of Changes

- Implements `TestflightConfigurationProcessor` `writeConfiguration`
- Makes `BetaTester` arrays non optional
- Completes `testflight sync pull`

# ⚠️ Items of Note

If we use our own FileSystem app model we can get rid of an unnecessary `!` in a subsequent PR and ensure we only write out what we care about

# 🧐🗒 Reviewer Notes

## 💁 Example

`asc testflight sync pull`

`asc testflight sync pull --path "./appConfig"`

`asc testflgith sync pull --filter-bundle-ids com.example.app`

## 🔨 How To Test

Run `testflight sync pull` with various options and ensure that it writes out the expected config

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
+++ b/Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
@@ -315,22 +315,29 @@ class AppStoreConnectService {
             }
         }
 
-        return try ListBetaTestersOperation(options:
-            .init(
+        let operation = ListBetaTestersOperation(
+            options: .init(
                 email: email,
                 firstName: firstName,
                 lastName: lastName,
                 inviteType: inviteType,
-                appIds: filterAppIds,
+                appIds: nil, // Specifying app ids in the API request can cause undefined behaviour
                 groupIds: groupIds,
                 sort: sort,
                 limit: limit,
                 relatedResourcesLimit: relatedResourcesLimit
             )
         )
-        .execute(with: requestor)
-        .await()
-        .map(Model.BetaTester.init)
+
+        var betaTesters = try operation.execute(with: requestor).await()
+
+        if filterAppIds.isEmpty == false {
+            betaTesters = betaTesters.filter { betaTester in
+                betaTester.apps?.first(where: { app in filterAppIds.contains(app.id) }) != nil
+            }
+        }
+
+        return betaTesters.map(Model.BetaTester.init)
     }
 
     func listBetaTestersForGroup(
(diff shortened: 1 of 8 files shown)


Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -315,22 +315,29 @@ class AppStoreConnectService {
...
                 relatedResourcesLimit: relatedResourcesLimit
             )
         )
-        .execute(with: requestor)
-        .await()
-        .map(Model.BetaTester.init)
+
+        var betaTesters = try operation.execute(with: requestor).await()
+
+        if filterAppIds.isEmpty == false {
```
DechengMa, line 334:
```suggestion
        if filterAppIds.isNotEmpty {
```

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -1,22 +1,83 @@
...
+            let headers = [BetaTester.CodingKeys.allCases.map(\.rawValue)]
+            let rows = testers.map { [$0.email, $0.firstName, $0.lastName] }
+            return headers + rows
+        }
+
+        let filenameForBetaGroup: (BetaGroup) -> String = { betaGroup in
+            return betaGroup.groupName
+                .components(separatedBy: CharacterSet(charactersIn: " *?:/\\."))
+                .joined(separator: "_")
+                + ".yml"
```
DechengMa, line 56: How about extracting the naming logic to be a var or helper in extension to `BetaGroup`?
```
extension FileSystem.BetaGroup {
    var filename: String { xxx }
}
```

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -1,22 +1,83 @@
...
+
+            config.betaGroups = (groupsByApp[app.id] ?? []).map { betaGroup in
+                FileSystem.BetaGroup(
+                    betaGroup: betaGroup,
+                    betaTesters: testers.filter { $0.betaGroups.map(\.id).contains(betaGroup.id) }
+                )
+            }
+
+            return config
+        }
```
DechengMa, line 40: I guess the `// Generate configuration` can be separated into a different function for easy testing purpose

DechengMa (approved):
Looks good, let's move forward

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
adamjcampbell: Potentially yes. Might be worth making some kind of top level file system service that the commands interact with and the processor might only need to be responsible for writing the files.

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
adamjcampbell: Since it's only used in one place I didn't see a reason at the moment to share it across the module

PR #228, opened 2020-09-01
Create FileSystem App model and throw if an app is missing a bundleId

adamjcampbell:
# 📝 Summary of Changes

- Adds FileSystem.App
- Removes a force unwrap
- Adds `ModelError` for when we need to throw errors from `FileSystem` related to malformed model data

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

Sync pull should remain unchanged

### DIFF
--- a/Sources/FileSystem/Model/App.swift
+++ b/Sources/FileSystem/Model/App.swift
@@ -1,0 +1,26 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Model
+
+struct App: Codable {
+
+    var id: String
+    var bundleId: String
+    var name: String?
+    var primaryLocale: String?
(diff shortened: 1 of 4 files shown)


DechengMa (approved):
LGTM

Sources/FileSystem/Model/App.swift
```diff
@@ -0,0 +1,26 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Model
+
+struct App: Codable {
+
+    let id: String
```
DechengMa, line 8: I wonder here why don't make id as `var` as well?

adamjcampbell: I'll amend it

PR #230, opened 2020-09-02
Read written configuration from disk

adamjcampbell:
# 📝 Summary of Changes

- Adds `TestFlightPushCommand`
- Adds `TestflightProgram`
- Adds `readConfiguration` to `TestflightConfigurationProcessor`

# ⚠️ Items of Note

Push is still unimplemented.

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

Run `asc testflight sync pull` followed by `asc testflight sync push`. Breakpoint in a development environment to check that the configuration has been formed correctly.

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
@@ -34,7 +34,8 @@ struct TestFlightPullCommand: CommonParsableCommand {
         let testers = try service.listBetaTesters(filterIdentifiers: identifiers, limit: 200)
         let groups = try service.listBetaGroups(filterIdentifiers: identifiers)
 
-        try FileSystem.writeConfiguration(apps: apps, testers: testers, groups: groups, to: outputPath)
+        let testflightProgram = TestflightProgram(apps: apps, testers: testers, groups: groups)
+        try FileSystem.writeTestflightConfiguration(program: testflightProgram, to: outputPath)
     }
 
 }
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
@@ -34,8 +34,8 @@ struct TestFlightPullCommand: CommonParsableCommand {
         let testers = try service.listBetaTesters(filterIdentifiers: identifiers, limit: 200)
         let groups = try service.listBetaGroups(filterIdentifiers: identifiers)
 
-        let testflightProgram = TestflightProgram(apps: apps, testers: testers, groups: groups)
-        try FileSystem.writeTestflightConfiguration(program: testflightProgram, to: outputPath)
+        let testflightProgram = TestFlightProgram(apps: apps, testers: testers, groups: groups)
+        try FileSystem.writeTestFlightConfiguration(program: testflightProgram, to: outputPath)
     }
 
 }
--- a/Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
+++ b/Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
@@ -54,28 +54,38 @@ struct TestflightConfigurationProcessor {
         }
     }
 
-    enum ReadError: LocalizedError {
-        case fileIsNotUTF8(filePath: String)
-
-        var errorDescription: String? {
-            return nil
-        }
-    }
-
     func readConfiguration() throws -> TestflightConfiguration {
         let folder = try Folder(path: path)
+        var configurations: [TestflightConfiguration.AppConfiguration] = []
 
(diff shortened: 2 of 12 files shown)


Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -54,4 +54,43 @@ struct TestflightConfigurationProcessor {
...
+                try BetaTester(
+                    email: record[BetaTester.CodingKeys.email.rawValue],
+                    firstName: record[BetaTester.CodingKeys.firstName.rawValue],
+                    lastName: record[BetaTester.CodingKeys.lastName.rawValue]
+                )
+            }
+        }
+
+        configuration.appConfigurations = try folder.subfolders.map { appFolder in
+            let appYAML = try appFolder.file(named: "app.yml").readAsString()
```
adamjcampbell, line 77: Common file names such as `"app.yml"` and `"beta-testers.csv"` will be extracted to constants to share between read and write

Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
```diff
@@ -34,7 +34,8 @@ struct TestFlightPullCommand: CommonParsableCommand {
         let testers = try service.listBetaTesters(filterIdentifiers: identifiers, limit: 200)
         let groups = try service.listBetaGroups(filterIdentifiers: identifiers)
 
-        try FileSystem.writeConfiguration(apps: apps, testers: testers, groups: groups, to: outputPath)
+        let testflightProgram = TestflightProgram(apps: apps, testers: testers, groups: groups)
```
DechengMa, line 37: Which one do we want to use, Test`F`light  Test`f`light? 🤔

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -54,4 +58,43 @@ struct TestflightConfigurationProcessor {
         }
     }
 
+    func readConfiguration() throws -> TestflightConfiguration {
+        let folder = try Folder(path: path)
+        var configuration = TestflightConfiguration()
+
+        let decodeBetaTesters: (Data) throws -> [BetaTester] = { data in
```
DechengMa, line 65: Just a suggestion:
```suggestion
extension Array where Element == BetaTester {
    init(data: Data) throws {
    }
}
```

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
```diff
@@ -54,4 +58,43 @@ struct TestflightConfigurationProcessor {
...
+        }
+
+        configuration.appConfigurations = try folder.subfolders.map { appFolder in
+            let appYAML = try appFolder.file(named: appYAMLName).readAsString()
+            let app = try YAMLDecoder().decode(from: appYAML) as App
+
+            var appConfiguration = TestflightConfiguration.AppConfiguration(app: app)
+
+            let testersFile = try appFolder.file(named: betaTestersCSVName)
+            appConfiguration.betaTesters = try decodeBetaTesters(try testersFile.read())
```
DechengMa, line 87: then here we can 
```suggestion
            appConfiguration.betaTesters = try [BetaTester].init(data: testersFile.read())
```

Sources/Model/TestflightProgram.swift
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+
+/// Aggregated model data representing the a Testflight Beta Program (Apps, Testers and Groups)
+public struct TestflightProgram {
```
DechengMa, line 6: This looks good now. Will these data be enough for our following comparison logic 🤔 ?

DechengMa (approved):
So far so good 👍

Sources/Model/TestflightProgram.swift
DechengMa: What I am worried the most is `BetaTester` coz it doesn't have `var id`

Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightPullCommand.swift
adamjcampbell: I guess we should just use `TestFlight` everywhere since it's the brand capitalisation

Sources/FileSystem/Configuration/TestflightConfigurationProcessor.swift
adamjcampbell: I believe this makes the code a lot less discoverable. I realise it's a common thing in this codebase, however I don't believe it's that common a practice. I also think there is merit in limiting the scope of functions, as in why should we expose this functionality more widely than the single place it is used?

Sources/Model/TestflightProgram.swift
adamjcampbell: We may need to add it in the PR that begins to form those comparisons. Better to leave it until it's required

PR #231, opened 2020-09-03
Convert TestflightConfiguration into TestflightProgram for comparison

adamjcampbell:
# 📝 Summary of Changes

- Add extension to `TestFlightProgram` to convert a `TestFlightConfiguration`
- Returns fully qualified `TestFlightProgram` from `readTestFlightConfiguration`

# ⚠️ Items of Note

This command is still under active development and throws an unimplemented error

# 🧐🗒 Reviewer Notes

## 🔨 How To Test

Run `asc testflight sync pull` followed by `asc testflight sync push`. Breakpoint in a development environment to check that the TestFlight Program has been formed correctly.


DechengMa (approved):
I've tested this locally and it works nice.
There is one thing that might be worth to take a look: when creating new group locally(a group file without ID inside), testers inside that group can't be read properly.

PR #233, opened 2020-09-07
Process TestflightProgramDifference

adamjcampbell:
# 📝 Summary of Changes

- Adds `TestflightProgramDifference`
- Calculates add and remove operations for beta groups and beta testers

# ⚠️ Items of Note

This is just a part of getting dry run for sync up and running. The command is still unimplemented and would be nice to make a testing set to gain confidence in how it operates.

# 🧐🗒 Reviewer Notes

## 💁 Example
```
$ swift run asc testflight sync push
Beta group named: testGroup2 will be added to app with bundleId: com.example
Beta Tester with email: example@gmail.com will be added to groups: testGroup2
Error: This command has not been implemented
```

## 🔨 How To Test

Add beta groups / beta testers to existing apps after using `asc testflight sync pull` to see if this behaves correctly

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
@@ -97,8 +97,8 @@ struct TestFlightProgramDifference {
                     .filter { !localTester.betaGroups.map(\.id).contains($0.id) }
                 let removeFromGroups = Change.removeBetaTesterFromGroups(localTester, groupsToRemove)
                 changes += groupsToRemove.isNotEmpty ? [removeFromGroups] : []
-            } else if remoteBetaGroups.isNotEmpty {
-                changes.append(.removeBetaTesterFromGroups(remoteTester, remoteBetaGroups))
+            } else if remoteApps.isNotEmpty {
+                changes.append(.removeBetaTesterFromApps(remoteTester, remoteApps))
             }
         }
 
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
@@ -1,0 +1,77 @@
+// Copyright 2020 Itty Bitty Apps Pty Ltd
+
+import Foundation
+import Model
+
+struct TestFlightProgramDifference {
+
+    enum Change {
+        case addBetaGroup(BetaGroup)
+        case removeBetaGroup(BetaGroup)
+        case addBetaTester(BetaTester, toBetaGroups: [BetaGroup])
+        case removeBetaTester(BetaTester, fromBetaGroups: [BetaGroup])
+
+        var description: String {
+
+            let operation: String = {
+                switch self {
+                case .addBetaGroup, .addBetaTester: return "added to"
+                case .removeBetaGroup, .removeBetaTester: return "removed from"
+                }
+            }()
+
+            switch self {
+            case .addBetaGroup(let betaGroup), .removeBetaGroup(let betaGroup):
+                let name = betaGroup.groupName ?? ""
+                let bundleId = betaGroup.app?.bundleId ?? ""
+
+                return "Beta group named: \(name) will be \(operation) app with bundleId: \(bundleId)"
+
+            case .addBetaTester(let betaTester, let betaGroups),
+                 .removeBetaTester(let betaTester, let betaGroups):
+                let email = betaTester.email ?? ""
+                let groupNames = betaGroups.map({ $0.groupName ?? "" }).joined(separator: ", ")
+
+                return "Beta Tester with email: \(email) will be \(operation) groups: \(groupNames)"
+            }
+        }
+    }
+
+    let changes: [Change]
+
+    init(local: TestFlightProgram, remote: TestFlightProgram) {
+        var changes: [Change] = []
+
+        // Groups
+        let localGroups = local.groups
+        let groupsToAdd: [BetaGroup] = localGroups.filter { $0.id == nil }
+        changes += groupsToAdd.map(Change.addBetaGroup)
+
+        let localGroupIds = localGroups.map(\.id)
+        let groupsToRemove: [BetaGroup] = remote.groups
+            .filter { group in localGroupIds.contains(group.id) == false }
+        changes += groupsToRemove.map(Change.removeBetaGroup)
+
+        // Testers
+        let newTesters = local.testers.filter { !remote.testers.map(\.email).contains($0.email) }
+        changes += newTesters.map { Change.addBetaTester($0, toBetaGroups: $0.betaGroups) }
+
+        for remoteTester in remote.testers {
+            if let localTester = local.testers.first(where: { $0.email == remoteTester.email }) {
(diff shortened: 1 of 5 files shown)


Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
```diff
@@ -0,0 +1,77 @@
...
+                let bundleId = betaGroup.app?.bundleId ?? ""
+
+                return "Beta group named: \(name) will be \(operation) app with bundleId: \(bundleId)"
+
+            case .addBetaTester(let betaTester, let betaGroups),
+                 .removeBetaTester(let betaTester, let betaGroups):
+                let email = betaTester.email ?? ""
+                let groupNames = betaGroups.map({ $0.groupName ?? "" }).joined(separator: ", ")
+
+                return "Beta Tester with email: \(email) will be \(operation) groups: \(groupNames)"
```
DechengMa, line 35: What do you think of adding `bundleId` in this one? Otherwise, when two apps have two groups with the same name, this will become confusing.
```suggestion
                return "Beta Tester with email: \(email) will be \(operation) groups: \(groupNames) in bundleId: \(bundleId)"
```

Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
```diff
@@ -0,0 +1,77 @@
...
+        let groupsToAdd: [BetaGroup] = localGroups.filter { $0.id == nil }
+        changes += groupsToAdd.map(Change.addBetaGroup)
+
+        let localGroupIds = localGroups.map(\.id)
+        let groupsToRemove: [BetaGroup] = remote.groups
+            .filter { group in localGroupIds.contains(group.id) == false }
+        changes += groupsToRemove.map(Change.removeBetaGroup)
+
+        // Testers
+        let newTesters = local.testers.filter { !remote.testers.map(\.email).contains($0.email) }
```
DechengMa, line 56: After some testing, I find that if a tester is already in Group A, when you also want to add it to group B, this function can't detect that.
 
There might be a scennario missing here? A tester is in server, but the group in that tester is different.
```suggestion
        let newTesters = local.testers.filter { 
            if !remote.testers.map(\.email).contains(tester.email) { return true }
            else if remote.testers.map(\.email).contains(tester.email) {
                // Check if they have the same group, If not return true
            }
            return false
        }
```

DechengMa (changes requested):

Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
adamjcampbell: @DechengMa can you please post an example so that I can try it myself?
Also this `newTesters`section was meant to find testers that don't currently exist. So I would want to modify the section of code underneath this I think

adamjcampbell: @DechengMa I believe I have fixed this edge case

Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
adamjcampbell: 🤔 for this one there one there is not one bundleId there could be groups here across many apps
So it would need to be a list of group names and bundle ids e.g.

"Beta Tester with email: example@bla.com will be added to group1 in com.example.app1, group2 in com.example.app2"

DechengMa:
Just found another edge case: 
when adding a tester to `beta-testers.csv`, but not add it to any group:
It will say `Beta Tester with email: foo@gmail.com will be added to groups: ` with an empty group list. Do we want to show `will be added to app: bundleId` or just skip in this case?

adamjcampbell:
> Just found another edge case: when adding a tester to beta-testers.csv, but not add it to any group

I'm a bit torn on this one. I'm thinking that it's not actually valid to do this. It should be added to at least one group when we read from file. So I think I'll throw an error in this case and state that it needs to be added to a group.

I should probably also error out in the case that it's added to a group but not the main testers file. What do you think?

DechengMa:
> > Just found another edge case: when adding a tester to beta-testers.csv, but not add it to any group
> 
> I'm a bit torn on this one. I'm thinking that it's not actually valid to do this. It should be added to at least one group when we read from file. So I think I'll throw an error in this case and state that it needs to be added to a group.
> 
> I should probably also error out in the case that it's added to a group but not the main testers file. What do you think?

Can throw an error for now as a reminder, I agree

adamjcampbell:
@DechengMa added an error for the no beta groups instance:

Example:
```
$ asc testflight sync push
Error: Tester with email: bob@jones.com being added to apps: iba.test2 has not been added to any beta groups
```

DechengMa:
> @DechengMa added an error for the no beta groups instance:
> 
> Example:
> 
> ```
> $ asc testflight sync push
> Error: Tester with email: bob@jones.com being added to apps: iba.test2 has not been added to any beta groups
> ```

Sorry I found this is not quite feasible. In my test account, I already have some existing testers in an App that has no been assign to any beta groups.
If I'd like to perform a `deleteTesterfromApp`, this will not allow me to do. 😢  Any thoughts?

adamjcampbell:
> Sorry I found this is not quite feasible. In my test account, I already have some existing testers in an App in server that has no been assign to any beta groups.
> If I'd like to perform a deleteTesterfromApp, this will not allow me to do. 😢 Any thoughts?

In this case should you not delete those testers from the beta-testers.csv file?

DechengMa:
> > Sorry I found this is not quite feasible. In my test account, I already have some existing testers in an App in server that has no been assign to any beta groups.
> > If I'd like to perform a deleteTesterfromApp, this will not allow me to do. 😢 Any thoughts?
> 
> In this case should you not delete those testers from the beta-testers.csv file?

If I delete those, in next sync pull, I guess they will come back again. 🤔

adamjcampbell:
> If I delete those, in next sync pull, I guess they will come back again. 🤔

I guess we do need to implement add to app and remove from app then not just groups

adamjcampbell:
@DechengMa pushed some changes covering more add and remove app cases.

This all behaves fine when you pull down the entire remote configuration but I think we'll need to pay attention to having the correct behaviour if you are only meant to change the apps which you have pulled down (with a filter) and not others

DechengMa:
> @DechengMa pushed some changes covering more add and remove app cases.
> 
> This all behaves fine when you pull down the entire remote configuration but I think we'll need to pay attention to having the correct behaviour if you are only meant to change the apps which you have pulled down (with a filter) and not others

I have tested these code and it works nice. Can merge when you happy 😃 .
I also have another minor suggestion you might want to consider: 
If we can have error been thrown when a tester is added to a group, but test is not existing in that the app, that will be 💯 
eg.
```
id: 123456-ae73-421c-ad07-123456
groupName: New Group 8
testers:
- example@google.com
```
and the `example@google.com` doesn't exist in `beta-testers.csv`. Currently, this will produce no result.

DechengMa (approved):

adamjcampbell:
@DechengMa I added that error you were looking for:
It should print out in the format:
`Error: Tester with email: foo@bar.com in beta group named: groupName for app: com.example.app is not included in the beta-testers.csv file`

PR #235, opened 2020-09-29
Change ListProfilesByBundleIdCommand to use fuzzy search

DechengMa:
Remove bundle id must be unique validation for `ListProfilesByBundleIdCommand`.
Input first part of the bundle ID will result in all the matched cases.

# 📝 Summary of Changes

- Removed bundle id `must be unique validation` for `ListProfilesByBundleIdCommand`.
- Return all the matched profiles for input bundle Id

# ⚠️ Items of Note

Inputted bundle Id string must be the first part of the bundle Id.

# 🧐🗒 Reviewer Notes

## 💁 Example

USAGE: 
```
asc profiles list-by-bundle [--api-issuer <uuid>] [--api-key-id <keyid>] [--csv] [--json] [--table] [--yaml] [--verbose] <bundle-id> [--limit <limit>] [--download-path <path>]
```

## 🔨 How To Test

```
swift run asc profiles list-by-bundle com --download-path ../../
```

### DIFF
--- a/Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
+++ b/Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
@@ -4,12 +4,12 @@ import ArgumentParser
 import AppStoreConnect_Swift_SDK
 import Foundation
 
-private typealias Filter = DownloadSalesAndTrendsReports.Filter
+public typealias Filter = DownloadSalesAndTrendsReports.Filter
 
 extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
-    private typealias AllCases = [Filter.Frequency]
+    public typealias AllCases = [Filter.Frequency]
 
-    private static var allCases: AllCases {
+    public static var allCases: AllCases {
         [.DAILY, .MONTHLY, .WEEKLY, .YEARLY]
     }
 
@@ -23,9 +23,9 @@ extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
 }
 
 extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
-    private typealias AllCases = [Filter.ReportType]
+    public typealias AllCases = [Filter.ReportType]
 
-    private static var allCases: AllCases {
+    public static var allCases: AllCases {
         [.SALES, .PRE_ORDER, .NEWSSTAND, .SUBSCRIPTION, .SUBSCRIPTION_EVENT, .SUBSCRIBER]
     }
 
@@ -39,9 +39,9 @@ extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
 }
 
 extension Filter.ReportSubType: ExpressibleByArgument, CustomStringConvertible {
-    private typealias AllCases = [Filter.ReportSubType]
+    public typealias AllCases = [Filter.ReportSubType]
 
-    private static var allCases: AllCases {
+    public static var allCases: AllCases {
         [.SUMMARY, .DETAILED, .OPT_IN]
     }
 
--- a/Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
+++ b/Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
@@ -9,7 +9,7 @@ private typealias Filter = DownloadSalesAndTrendsReports.Filter
 extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.Frequency]
 
-    public static var allCases: AllCases {
+    private static var allCases: AllCases {
         [.DAILY, .MONTHLY, .WEEKLY, .YEARLY]
     }
 
@@ -25,7 +25,7 @@ extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
 extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.ReportType]
 
-    public static var allCases: AllCases {
+    private static var allCases: AllCases {
         [.SALES, .PRE_ORDER, .NEWSSTAND, .SUBSCRIPTION, .SUBSCRIPTION_EVENT, .SUBSCRIBER]
     }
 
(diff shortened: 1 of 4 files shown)


Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -9,7 +9,7 @@ private typealias Filter = DownloadSalesAndTrendsReports.Filter
 extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.Frequency]
 
-    public static var allCases: AllCases {
+    private static var allCases: AllCases {
```
orj, line 12: This seems unrelated?

DechengMa: > This seems unrelated?

This is for CLI project to be running under Xcode 12. I don't know why it's required but Xcode wouldn't compile without this

orj: Right.  Yeah. I noticed earlier actually. Was going to ask.

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -9,7 +9,7 @@ private typealias Filter = DownloadSalesAndTrendsReports.Filter
 extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.Frequency]
```
orj, line 10: You could probably fix this by changing:
```suggestion
    public typealias AllCases = [Filter.Frequency]
```

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
orj: If you changed the type alias to public you could leave the other stuff public.

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -25,7 +25,7 @@ extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
 extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.ReportType]
```
orj, line 26:
```suggestion
    public typealias AllCases = [Filter.ReportType]
```

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -25,7 +25,7 @@ extension Filter.Frequency: ExpressibleByArgument, CustomStringConvertible {
...
 
-    public static var allCases: AllCases {
+    private static var allCases: AllCases {
```
orj, line 28:
```suggestion
    public static var allCases: AllCases {
```

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -41,7 +41,7 @@ extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
 extension Filter.ReportSubType: ExpressibleByArgument, CustomStringConvertible {
     private typealias AllCases = [Filter.ReportSubType]
```
orj, line 42:
```suggestion
    public typealias AllCases = [Filter.ReportSubType]
```

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
```diff
@@ -41,7 +41,7 @@ extension Filter.ReportType: ExpressibleByArgument, CustomStringConvertible {
...
 
-    public static var allCases: AllCases {
+    private static var allCases: AllCases {
```
orj, line 44:
```suggestion
    public static var allCases: AllCases {
```

orj (commented):
The `typelias` is what should be public perhaps.

Sources/AppStoreConnectCLI/Model/API/Sales.Filter+ExpressibleByArgument.swift
DechengMa: just changed it to public and pushed ✅

orj:
If you could update the documentation on the `list-by-bundle` command too to reflect the change in behaviour. That would be great.

DechengMa:
> If you could update the documentation on the `list-by-bundle` command too to reflect the change in behaviour. That would be great.

Cool, will do, what do you think of this?
```
abstract: "Find and list provisioning profiles and download their data by a fuzzy search for bundleId.",
```

orj:
I think the behaviour should be slightly different too.  I think it should only match the beginning of the bundle ID.  We'll have to do some local filtering for that.

orj:
I think the abstract is ok as is.  I think we should add further discussion to the `bundleId` argument:

```
@Argument(help: "The bundle ID of an application. (eg. com.example.app)",
              discussion: "Note that all provisioning profiles for bundle identifiers beginning with this string will be fetched.")
    var bundleId: String
```

Sources/AppStoreConnectCLI/Services/AppStoreConnectService.swift
```diff
@@ -703,19 +703,23 @@ class AppStoreConnectService {
...
-            options: .init(bundleId: bundleId)
+        let bundleIdResourceIds = try ListBundleIdsOperation(
+            options: .init(identifiers: [bundleId], names: [], platforms: [], seedIds: [], limit: nil)
         )
-            .execute(with: requestor)
-            .await()
-            .id
+        .execute(with: requestor)
+        .await()
+        .filter { ($0.attributes?.identifier?.starts(with: bundleId) ?? false) }
```
orj, line 711: Nice 👍

orj (approved):
Nice work @DechengMa. Thanks.

PR #238, opened 2020-10-08
Don't send deviceIds when provision type is AppStore

ptmt:
# 📝 Summary of Changes

When you create a AppStore provisioning profile, you don't need to send deviceIds.

### DIFF
--- a/Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
@@ -32,8 +32,6 @@ struct CreateProfileOperation: APIOperation {
             .map { $0.data }
             .eraseToAnyPublisher()
     }
-    
-    
 
 }
 
--- a/Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
+++ b/Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
@@ -17,12 +17,15 @@ struct CreateProfileOperation: APIOperation {
     private let endpoint: APIEndpoint<ProfileResponse>
 
     init(options: Options) {
+        let shouldOmitDeviceIds = [.macAppStore, .iOSAppStore, .tvOSAppStore]
+            .contains(options.profileType)
+        
         endpoint = APIEndpoint.create(
             profileWithId: options.bundleId,
             name: options.name,
             profileType: options.profileType,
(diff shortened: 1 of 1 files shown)


Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
```diff
@@ -32,5 +32,13 @@ struct CreateProfileOperation: APIOperation {
...
+    
+    
 
 }
+
+fileprivate extension ProfileType {
+    var areDeviceIdsRequired: Bool {
+        return !self.rawValue.contains("_APP_STORE")
+    }
+}
```
adamjcampbell, line 44: This check might be more obvious if it was in the initializer rather than a private extension.
Also the rawValue comparison requires knowledge of how the profile type is formed. For a reader of the code checking against enums might feel less awkward.

I'm thinking something like:
```swift
        let shouldOmitDeviceIds = [.macAppStore, .iOSAppStore, .tvOSAppStore]
            .contains(options.profileType)

        endpoint = APIEndpoint.create(
            profileWithId: options.bundleId,
            name: options.name,
            profileType: options.profileType,
            certificateIds: options.certificateIds,
            deviceIds: shouldOmitDeviceIds ? options.deviceIds : []
        )
```

What do you think?

Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
```diff
@@ -32,5 +32,13 @@ struct CreateProfileOperation: APIOperation {
             .map { $0.data }
             .eraseToAnyPublisher()
     }
+    
+
```
adamjcampbell, line 36: The added whitespace here is triggering `swiftlint` if you open the workspace instead of the package you should be able to see the warnings more easily.
```suggestion
```

adamjcampbell (changes requested):
Thank you for this change, are you able to elaborate on the issues this was causing? Or is it coming from a place of correctness / only sending what is neccesary

Sources/AppStoreConnectCLI/Services/Operations/CreateProfileOperation.swift
ptmt: Thanks, this is much better!

adamjcampbell:
Thanks for the update @ptmt once this is up to date with master I'll be happy to merge it 👍

adamjcampbell (approved):

PR #239, opened 2020-10-08
Handle situation where you have nested bundle id

ptmt:
# 📝 Summary of Changes

Imagine, If you have an app `com.mycompany.myapp` and say a share extension with the `com.mycompany.myapp.share` bundle id you will get a `bundleIdNotUnique` exception. 
This change fixes that.


DechengMa (approved):
@ptmt Thanks for your contribution!

PR #241, opened 2020-10-12
Adding data set validation for sync command

DechengMa:
Adding data set validation for sync command

# 📝 Summary of Changes

- TODO

# ⚠️ Items of Note

TODO

# 🧐🗒 Reviewer Notes

## 💁 Example

TODO

## 🔨 How To Test

TODO

### DIFF
--- a/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
+++ b/Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
@@ -52,9 +52,20 @@ struct TestFlightProgramDifference {
         }
     }
 
+    enum Error: LocalizedError, Equatable {
+        case duplicateTesters(email: String)
+
+        var errorDescription: String? {
+            switch self {
+            case .duplicateTesters(let email):
+                return "There are two beta testers with the same email '\(email)' exists in your account, please clean that up before continue."
+            }
+        }
+    }
+
     let changes: [Change]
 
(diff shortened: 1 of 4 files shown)


Sources/AppStoreConnectCLI/Commands/TestFlight/Sync/TestFlightProgramDifference.swift
```diff
@@ -52,9 +52,20 @@ struct TestFlightProgramDifference {
         }
     }
 
+    enum Error: LocalizedError, Equatable {
```
adamjcampbell, line 55: There's an argument to be made around not adding equatable conformance in the implementation if you only use it for testing. Not sure how I feel about it though

DechengMa: > There's an argument to be made around not adding equatable conformance in the implementation if you only use it for testing. Not sure how I feel about it though

Any ideas for removing the Equatable in this case? 🤔

adamjcampbell: If you wanted to you could replace the instances in testing with something like:
```swift
        XCTAssertThrowsError(try TestFlightProgramDifference(local: local, remote: remote)) { error in
            guard case TestFlightProgramDifference.Error.duplicateTesters(let email) = error else {
                return XCTFail()
            }

            XCTAssertEqual(email, "foo@gmail.com")
        }
```

PR #246, opened 2021-01-30
Change Appstore Connect SDK dependency 

mattnydam:
Updating the Appstore connect SDK to point to origin. 

From what I can tell, all changes stemming from the forked SDK have been merged upstream. If this is not the case or there is still a reason to keep using a fork, we can close this PR and instead keep the fork up to date, corresponding [PR here](https://github.com/ittybittyapps/appstoreconnect-swift-sdk/pull/8)

This initially generated a warning in `Platform+ExpressibleByArgument.swift` as CaseIterable is now declared in `Platform.swift`

Build and run succeeds and all tests pass. 

# 📝 Summary of Changes

- Change Appstore Connect SDK dependency to the [origin repo](https://github.com/AvdLee/appstoreconnect-swift-sdk)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,11 +3,11 @@
     "pins": [
       {
         "package": "AppStoreConnect-Swift-SDK",
-        "repositoryURL": "https://github.com/ittybittyapps/appstoreconnect-swift-sdk.git",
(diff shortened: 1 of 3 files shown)


orj (approved):
We can flick over to origin for the moment.  We do maintenance of the upstream in our fork.  So if we need to make changes to the SDK in future we can flick back to our fork.

